PR #18, opened 2022-06-23
enable multi-arch build

andysim3d:


bwplotka:
Thanks! Sorry for massive lag - our dockerfiles changed since then - let me know if this is still required, we can reopen

PR #20, opened 2022-11-17
Export exemplars to GCM

realschwa:
This PR exports a map[storage.SeriesRef]record.RefExemplar to the export function in GMP. See https://github.com/GoogleCloudPlatform/prometheus-engine/pull/308

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v63.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.25
 	github.com/Azure/go-autorest/autorest/adal v0.9.18
-	github.com/GoogleCloudPlatform/prometheus-engine v0.4.1
+	github.com/GoogleCloudPlatform/prometheus-engine v0.6.0-rc.0
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
 	github.com/aws/aws-sdk-go v1.43.31
 	github.com/cespare/xxhash/v2 v2.1.2
@@ -77,7 +77,7 @@ require (
 	k8s.io/apimachinery v0.23.5
 	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/klog v1.0.0
-	k8s.io/klog/v2 v2.40.1
+	k8s.io/klog/v2 v2.80.1
 )
 
 replace (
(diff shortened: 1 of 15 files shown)


pintohutch:
@realschwa let me know when this is ready for review (i.e. tag me to review it)

tsdb/agent/db.go
```diff
@@ -799,13 +803,14 @@ func (a *appender) AppendExemplar(ref storage.SeriesRef, l labels.Labels, e exem
 			return 0, storage.ErrExemplarLabelLength
 		}
 	}
-
-	a.pendingExamplars = append(a.pendingExamplars, record.RefExemplar{
+	refE := record.RefExemplar{
```
pintohutch: nit: `refEx` or even just `ref`?

tsdb/head_append.go
```diff
@@ -444,14 +444,21 @@ func (a *headAppender) Commit() (err error) {
...
 
 	// No errors logging to WAL, so pass the exemplars along to the in memory storage.
+	exportExemplars := make(map[storage.SeriesRef]record.RefExemplar, len(a.exemplars))
 	for _, e := range a.exemplars {
 		s := a.head.series.getByID(chunks.HeadSeriesRef(e.ref))
 		// We don't instrument exemplar appends here, all is instrumented by storage.
 		if err := a.head.exemplars.AddExemplar(s.lset, e.exemplar); err != nil {
-			if err == storage.ErrOutOfOrderExemplar {
-				continue
+			if err != storage.ErrOutOfOrderExemplar {
```
pintohutch: Why did you change this logic around?

In the interest of keeping a lean fork, we'd like to change as little as possible.

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v63.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.25
 	github.com/Azure/go-autorest/autorest/adal v0.9.18
-	github.com/GoogleCloudPlatform/prometheus-engine v0.4.1
+	github.com/GoogleCloudPlatform/prometheus-engine v0.5.1-0.20221123211114-163beef5eba8
```
pintohutch: The main reason I can think of for depending on a tagged version is to ensure the user-agent injects the proper prometheus-engine export library used. We should probably wait to tag it there before doing the final cut here.

tsdb/head_append.go
realschwa: I think I switched this around to account for a special case that turned out not to be so special. I can switch it back around

tsdb/agent/db.go
realschwa: `refEx`

go.mod
pintohutch: @realschwa can you bump this to point to https://github.com/GoogleCloudPlatform/prometheus-engine/releases/tag/v0.6.0-rc.0?

realschwa: @pintohutch done

pintohutch (approved):

PR #23, opened 2023-01-09
version bump

pintohutch:


realschwa (approved):
We could also just remove the job -- I guess it's okay to keep it though as long as it's not firing on PR

PR #50, opened 2023-04-21
Disable unused lock threads cron

TheSpiritXIII:
https://github.com/GoogleCloudPlatform/prometheus/actions?query=branch%3Arelease-2.35.0-gmp

It's very noisy and does nothing for us. It just runs this bot:

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,9 +1,6 @@
 name: 'Lock Threads'
 
-on:
-  schedule:
-    - cron: '13 23 * * *'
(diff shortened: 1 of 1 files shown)


pintohutch:
Where's the noise you're seeing for this?

pintohutch:
Looks like this started about a [month ago](https://github.com/GoogleCloudPlatform/prometheus/actions/runs/4473824188) due to switching the release branch as the "primary" branch.

This is just noise from our perspective so LGTM

pintohutch (approved):

PR #51, opened 2023-04-24
Security patch for CVE-2022-41723

TheSpiritXIII:
Update x/net to 0.7 to address CVE-2022-41723.

x/net 0.7 requires Golang 1.17 minimum. I decided to upgrade to Go 1.18 because Golang 1.17 does not support `go mod tidy` without a flag indicating to be backwards or forwards compatible. We already compile with Golang 1.20 so this should be okay.


pintohutch:
Hrm - what are your thoughts on upgrading to Prometheus 2.43 from your [branch](https://github.com/GoogleCloudPlatform/prometheus/tree/TheSpiritXIII/release-2.43.0-gmp)?

It looks like this was addressed upstream in https://github.com/prometheus/prometheus/commit/4126985908ebb08a0e93021ef2476fd77731227e and available on 2.43.0.

TheSpiritXIII:
The problem with Prometheus 2.43 is that after 6 weeks (the Prometheus dev cycle length), 2.43 will not be getting further vulnerability fixes. Personally, I feel like Prometheus 2.37 makes more sense because it is an LTS release [1] and it has this vulnerability fix already too. 2.37 has less of a chance of breaking a customer's Prometheus deployment because it's not as large of a jump.

[1] https://prometheus.io/docs/introduction/release-cycle/

pintohutch:
Oh that works too. Do you have a link you can point me to where you found that 2.37 was patched with this fix?

TheSpiritXIII:
I have a 2.37 branch: https://github.com/GoogleCloudPlatform/prometheus/tree/TheSpiritXIII/release-2.37.6-gmp

From a git blame the version bump was done here: https://github.com/GoogleCloudPlatform/prometheus/commit/607fcd5266ee2d77884afcc7eaac3fcbc39ef25a

pintohutch:
Cool - I vote we use that

bwplotka:
👍🏽

bwplotka (approved):

pintohutch (approved):

TheSpiritXIII:
Deciding to merge this because:
1. 2.37 has an associated prometheus-engine change: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/468
2. 2.37 pulls new dependencies (whose licenses must be scanned).

bwplotka:
Let's go!

PR #64, opened 2023-06-27
incorporate cluster label fix

pintohutch:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v63.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.25
 	github.com/Azure/go-autorest/autorest/adal v0.9.18
-	github.com/GoogleCloudPlatform/prometheus-engine v0.6.2-rc.0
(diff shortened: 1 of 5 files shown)


pintohutch:
friendly ping here @bwplotka @TheSpiritXIII :)

bwplotka (commented):
Thanks!

This incorporates export changes to 2.35 but also updates 5 irrelevant go modules. Ideally we don't so this commit is self-contained (I doubt export module upgrade requires those).

I would recommend changing go.mod of github.com/GoogleCloudPlatform/prometheus-engine manually to v0.6.4-rc.0 and doing `go mod tidy`. This will pull what's needed but not touch any other dep.

pintohutch:
I think this is happening because the `release-0.6` branch also includes [changes](https://github.com/GoogleCloudPlatform/prometheus-engine/commit/46b559329d53c98d50d861e538b70c1d5e1850a7) to the underlying k8s library.

Lemme open up a release branch just for `release-0.6.2` to minimize the net changes.

bwplotka (approved):
Thanks!

PR #65, opened 2023-06-29
handle potential NPE

pintohutch:


bwplotka (approved):
Thanks! 

This will hit https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/series_cache.go#L341 which I assume will just log error and increment err stats.

pintohutch:
Yep exactly, specifically [this](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.7.0/pkg/export/transform.go#L393) metric will be incremented.

@TheSpiritXIII PTAL as it is merging into your dev branch 😄

PR #72, opened 2023-07-10
Commenting on the PR with prombench's status

saketjajoo:
Right after a "verified" user triggers a benchmark run, a message would be posted on the PR as a comment about the status of the prombench trigger along with the dashboard's link where the results can be viewed.


bwplotka (approved):
Thanks!

PR #74, opened 2023-07-13
Use structured logging

maxamins:
GKE is showing info logs as errors

This is by design
https://cloud.google.com/stackdriver/docs/solutions/gke/managing-logs#best_practices Use structured errors instead.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -214,6 +214,8 @@ func main() {
 		oldFlagRetentionDuration model.Duration
 		newFlagRetentionDuration model.Duration
 	)
+	af := promlog.AllowedFormat{}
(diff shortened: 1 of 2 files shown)


maxamins:
Using `log.NewJSONLogger(log.NewSyncWriter(os.Stderr))` displays logs correctly. 

image

This is how it looks currently:
image

pintohutch:
Drive-by, but what about just playing with the built-in logging flags?

AFAICT this does the same thing as `--log.format=json`, but I could be wrong.

image

maxamins:
@pintohutch those are exactly my thoughts. This PR confirms passing `--log.format=json` will solve the GKE logging issue. This was a quick testing method.

PR #75, opened 2023-07-17
Updating the GitHub Action File

saketjajoo:
This PR updates the GH Actions file to have the following changes:
- Renaming the steps in the workflow to make it more meaningful (code readability).
- Disabling prombench trigger on release cut (prombench now works only upon commenting on a PR).
- Adding comments to a PR when a prombench run is cancelled.
- Adding labels to a PR when a prombench run is triggered.
- Fixing the docker image URLs to point to the GHCR repo instead of a personal dockerhub account.

### DIFF
--- a/.github/workflows/prombench-gmp-collector.yml
+++ b/.github/workflows/prombench-gmp-collector.yml
@@ -33,8 +33,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-      contents: read
-      packages: read
     steps:
 
       - name: 'User Verification'
@@ -134,7 +132,7 @@ jobs:
           curl --silent -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_TOKEN" \
             -H "X-GitHub-Api-Version: 2022-11-28" \
             https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/issues/${PR_NUMBER}/labels \
-            -d '{"labels":["prombench"]}' || true
+            -d '{"labels":["prombench"]}'
       
       - name: 'Check if the cluster exists'
         run: |
--- a/.github/workflows/prombench-gmp-collector.yml
+++ b/.github/workflows/prombench-gmp-collector.yml
@@ -33,6 +33,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
+      contents: read
+      packages: read
     steps:
 
       - name: 'User Verification'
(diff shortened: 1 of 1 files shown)


.github/workflows/prombench-gmp-collector.yml
```diff
@@ -1,9 +1,9 @@
 name: "Run Prombench"
 
 on:
-  push:
-    tags:
-      - 'v*-gmp.*'
+  # push:
```
maxamins: Delete this. If needed we can figure out how to do it.

.github/workflows/prombench-gmp-collector.yml
```diff
@@ -36,9 +36,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
+      contents: read
```
maxamins: Why do you need these permissions?

.github/workflows/prombench-gmp-collector.yml
```diff
@@ -70,7 +72,7 @@ jobs:
         uses: 'google-github-actions/setup-gcloud@v1'
 
       
-      - name: check_benchmark_trigger_action
+      - name: 'Set environment variables based on the triggered action'
         run: |
           GMP_COLLECTOR_TAG=$(echo ${{ github.event.comment.body }} | head -n 1 | awk '{ print $2 }')
```
maxamins: This is quite confusing. How about `COLLECTOR_TAG_OR_CANCEL`

.github/workflows/prombench-gmp-collector.yml
```diff
@@ -126,9 +128,18 @@ jobs:
...
+          
+      - name: 'Adding the prombench label to the PR'
+        if: env.CANCEL_BENCHMARK == 'false'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl --silent -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_TOKEN" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/issues/${PR_NUMBER}/labels \
+            -d '{"labels":["prombench"]}' || true
```
maxamins: what is `|| true` doing?

maxamins (commented):
Good job!

.github/workflows/prombench-gmp-collector.yml
saketjajoo: This is done to only have the necessary permissions for the `GITHUB_TOKEN`. The [default permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) are not required as the token is only used for commenting on a PR and reading docker images.
Ref: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

.github/workflows/prombench-gmp-collector.yml
saketjajoo: It's just to ensure that even if the command fails for some reason, the workflow does not stop. If the add label command fails, the exit code is non zero which will stop the remaining steps in the workflow. The `|| true ` ensures that the exit code is always `0`.

maxamins: Wouldn't you need it for the comment in line 122.

maxamins: Let's add failure resistance when we migrate from curl to gh

.github/workflows/prombench-gmp-collector.yml
maxamins: Let's add this section if end up having permission issues.

maxamins (approved):

maxamins:
Thanks!

PR #76, opened 2023-07-18
Adding GITHUB_TOKEN and its associated permissions

saketjajoo:
Adding GITHUB_TOKEN and its associated permissions so that Actions can authenticate to GHCR and pull the image.


maxamins (approved):
Nice!

PR #77, opened 2023-07-25
Migrating prombench trigger from GitHub Actions to Google Cloudbuild.

saketjajoo:
This migration is done to use GCP's cloudbuild service instead of GitHub Actions because of security implications, internal organization policy and the ease of implementation (to a certain extent). Using cloudbuild eliminates the need to store sensitive credentials in GitHub's environment.


.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
+# https://cloud.google.com/build/docs/build-config-file-schema#options
+options:
+  env:
+  - 'CANCEL_BENCHMARK=false'
+  - 'CLUSTER_NAME=prombench'
+  - 'CREATE_CLUSTER=true'
+  - 'DASHBOARD_ID=$_DASHBOARD_ID'
+  - 'DOMAIN_NAME='
+  - 'GITHUB_STATUS_TARGET_URL='
+  - 'GKE_PROJECT_ID=$_GKE_PROJECT_ID'
```
maxamins: can we use default substitution of $PROJECT_ID https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_default_substitutions

reduces the need to add it as a trigger substitution variable.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
+# https://cloud.google.com/build/docs/build-config-file-schema#options
+options:
+  env:
+  - 'CANCEL_BENCHMARK=false'
+  - 'CLUSTER_NAME=prombench'
+  - 'CREATE_CLUSTER=true'
+  - 'DASHBOARD_ID=$_DASHBOARD_ID'
+  - 'DOMAIN_NAME='
+  - 'GITHUB_STATUS_TARGET_URL='
```
maxamins: not used

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  - 'GKE_PROJECT_ID=$_GKE_PROJECT_ID'
+  - 'GRAFANA_ADMIN_PASSWORD=admin'
+  - 'LAST_COMMIT_SHA='
```
maxamins: not used

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  - 'OAUTH_TOKEN='
+  - 'PROMBENCH_IMAGE=$_PROMBENCH_IMAGE'
+  - 'PROMETHEUS_IMAGE_REPOSITORY=gke.gcr.io/prometheus-engine/prometheus'
+  - 'PROVIDER=gke'
+  - 'SERVICEACCOUNT_CLIENT_EMAIL=$_SERVICEACCOUNT_CLIENT_EMAIL'
+  - 'WH_SECRET='
+  - 'ZONE=us-east4-a'
+  substitutionOption: ALLOW_LOOSE
```
maxamins: why are we skipping this?

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  _COMMENTER_ROLE: $(body.comment.author_association) # OWNER / MEMBER / etc. (https://docs.github.com/en/graphql/reference/enums#commentauthorassociation)
+  _GIT_REPO: $(body.repository.full_name) # GoogleCloudPlatform/prometheus
+  _PR_COMMENT_ACTION: $(body.action)
+  _PR_COMMENT_TEXT: $(body.comment.body)
+  _PR_NUMBER: $(body.issue.number)
+
+# https://cloud.google.com/build/docs/build-config-file-schema#availablesecrets
+availableSecrets:
+  secretManager:
+  - versionName: projects/${PROJECT_NUMBER}/secrets/prombench_service_account_adc_json/versions/1
```
maxamins: Do we want to expose our file name here? 

also I see versions/latest in the guide. Should we use latest in stead of 1?

ttps://cloud.google.com/build/docs/securing-builds/use-secrets

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          "$${URL}" \
+          -d '{"body": "'"$${BODY}"'"}'
+      }
+
+      # Checking the action on the PR (benchmark should trigger only when a new comment is created)
+      # Available actions are: created, deleted and edited -- https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads
+      if ! [ "${_PR_COMMENT_ACTION}" == "created" ]
+      then
+        echo "The PR comment must be created (not edited or deleted) to trigger the benchmark."
```
maxamins: Need to have a graceful exit. Maybe a global variable called exit that you can check between steps.

https://cloud.google.com/build/docs/configuring-builds/pass-data-between-steps

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  dynamicSubstitutions: true
+
+# https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values
+substitutions:
+  _COMMENTER: $(body.comment.user.login)
+  _COMMENTER_ROLE: $(body.comment.author_association) # OWNER / MEMBER / etc. (https://docs.github.com/en/graphql/reference/enums#commentauthorassociation)
+  _GIT_REPO: $(body.repository.full_name) # GoogleCloudPlatform/prometheus
+  _PR_COMMENT_ACTION: $(body.action)
+  _PR_COMMENT_TEXT: $(body.comment.body)
+  _PR_NUMBER: $(body.issue.number)
```
maxamins: _PR_NUMBER and _GIT_REPO are available in default subs https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_default_substitutions

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+
+        curl --silent -L -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer $${GH_PAT}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          "$${URL}" \
+          -d '{"body": "'"$${BODY}"'"}'
+      }
+
+      # Checking the action on the PR (benchmark should trigger only when a new comment is created)
```
maxamins: nit: combine line 69-89 into one If using OR. return log `Requirements not met to trigger prom bench because ...`

will make code simpler to read

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+      then
+
+        GITHUB_ORG=$(echo "${_GIT_REPO}" | cut -d'/' -f1)
+        GITHUB_REPO=$(echo "${_GIT_REPO}" | cut -d'/' -f2)
+        export GITHUB_ORG="$${GITHUB_ORG}" && export GITHUB_REPO="$${GITHUB_REPO}"
+
+        COMMENT_MSG="${_COMMENTER} does not have admin or write privileges. Therefore cannot execute the benchmark command."
+        GITHUB_URL="https://api.github.com/repos/$${GITHUB_ORG}/$${GITHUB_REPO}/issues/${_PR_NUMBER}/comments"
+        echo "$${COMMENT_MSG}"
+        post_to_github "$${GITHUB_URL}" "$${COMMENT_MSG}"
```
maxamins: Just put the curl command here since it's called once. It will use less lines 

```
        curl --silent -L -X POST \
          -H "Accept: application/vnd.github+json" \
          -H "Authorization: Bearer $${GH_PAT}" \
          -H "X-GitHub-Api-Version: 2022-11-28" \
          "$${URL}" \
          -d '{"body": "'"$${BODY}"'"}'
```

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+        then
+          
+          echo "Cancelling the benchmark run for PR #${_PR_NUMBER}."
+          echo "true" > /workspace/CANCEL_BENCHMARK
+
+        else
+          
+          echo "GMP Collector Tag: $${COLLECTOR_TAG_OR_CANCEL}"
+
+          if [[ "$${COLLECTOR_TAG_OR_CANCEL}" == '' ]]
```
maxamins: Let's simplify this code a bit. Let's avoid nested ifs and repeated code like line 128 and 136. This will improve readability. 

How about this pseudo code. 
If there is a tag check if it exists
if no tag or tag doesn't exist  fetch latest 
else: use passed tag

```
	GCLOUD_TAG = ''
	if [[ "$${COLLECTOR_TAG_OR_CANCEL}" != '' ]]
	then
		GCLOUD_TAG=$(gcloud container images list-tags --filter="tags:$${COLLECTOR_TAG_OR_CANCEL}" --format=json $${PROMETHEUS_IMAGE_REPOSITORY})
	fi
	if [[ "$${GCLOUD_TAG}" == "[]" ]] || [[ "$${COLLECTOR_TAG_OR_CANCEL}" == '' ]]
		echo "Since the collector version is missing, the latest one would be used."
		DEFAULT_COLLECTOR_TAG=$(gcloud container images list-tags "$${PROMETHEUS_IMAGE_REPOSITORY}" --sort-by=UPDATE_TIME,TAGS | tail -1 | awk '{ print $2 }')
		echo "$${DEFAULT_COLLECTOR_TAG}" > /workspace/PROMETHEUS_IMAGE_VERSION
		export PROMETHEUS_IMAGE_VERSION="$${DEFAULT_COLLECTOR_TAG}"
	else
              echo "$${COLLECTOR_TAG_OR_CANCEL}" > /workspace/PROMETHEUS_IMAGE_VERSION
              export PROMETHEUS_IMAGE_VERSION="$${COLLECTOR_TAG_OR_CANCEL}"
         fi 
```

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  - id: "Checking cluster's existance"
+    name: gcr.io/cloud-builders/gcloud
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        if $(gcloud container clusters describe "$${CLUSTER_NAME}" --project="$${GKE_PROJECT_ID}" --location="$${ZONE}" >/dev/null 2>&1)
+        then
+          CREATE_CLUSTER="false"
+          echo "$${CREATE_CLUSTER}" > /workspace/CREATE_CLUSTER
```
maxamins: can you inline echo "false" > /workspace/CREATE_CLUSTER since `CREATE_CLUSTER` is not used anywhere else, no need for a variable.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+          echo "The cluster (name: $${CLUSTER_NAME}) already exists. Therefore, skipping cluster creation."
+        else
+          CREATE_CLUSTER="true"
+          echo "$${CREATE_CLUSTER}" > /workspace/CREATE_CLUSTER
```
maxamins: same comment as 158

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+          echo "The cluster (name: $${CLUSTER_NAME}) does not exist, the cluster will get created before creating the nodes and triggering the benchmark process."
+        fi
+
+  - id: "Setting gcloud credentials for triggering the benchmark"
```
maxamins, line 154: can we move this before any gcloud command is run? this may cause issues for other projects in the future.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+    name: gcr.io/cloud-builders/gcloud
+    secretEnv: ['ADC']
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        echo "$${ADC}" > /workspace/adc.json
+
+  - id: "Triggering Benchmark Action"
```
maxamins: can you break this into steps? I think it will improve readability. check go/tott/674 as a ref.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+          ENV_VAR_NAME="$1"
+          if [ -f "/workspace/$${ENV_VAR_NAME}" ]
+          then
+            export "$${ENV_VAR_NAME}=$(cat /workspace/$${ENV_VAR_NAME})"
+          fi
+        }
+
+        # All the environment variables created or updated in any of the previous steps must be read
+        # from the local file and exported in this step.
+        get_or_set_env_var "CANCEL_BENCHMARK";
```
maxamins: can inline each export "$${ENV_VAR_NAME}=$(cat /workspace/$${ENV_VAR_NAME})"

can create each file `Setting environment variables` step to avoid the check on line 204

maxamins (commented):
Added some comments to improve readability and reliability. Let me know your thoughts are.

.cloudbuild/prombench.yaml
saketjajoo: This is used in the [comment-monitor](https://github.com/GoogleCloudPlatform/prometheus-test-infra/blob/master/tools/commentMonitor/client.go#L105). Although we don't use comment monitor, the deployment may fail if the comment-monitor fails to start up. Hence, thought to keep it, but have it empty.

.cloudbuild/prombench.yaml
saketjajoo: Having the `ALLOW_LOOSE` parameter means the build does not return an error if there's a missing substitution variable or a missing substitution.

.cloudbuild/prombench.yaml
saketjajoo: Since the current set-up is not based on the [GitHub repo](https://cloud.google.com/build/docs/automating-builds/github/connect-repo-github?generation=2nd-gen), the variables get substituted at runtime. Once this PR gets merged, we can change these variables to use the default available substitutions (maybe as a next iteration).

.cloudbuild/prombench.yaml
saketjajoo: Currently, the secrets in the secrets manager are versioned. Maybe we can use [version aliases](https://cloud.google.com/secret-manager/docs/assign-alias-to-secret-version) so that when secrets are rotated, the alias can point to the newer version and the "latest" version need not be updated in code.

.cloudbuild/prombench.yaml
saketjajoo: Having multiple checks instead of combining them into a single `if` gives us granularity while logging. We can check the logs to understand why something might have gone wrong if something goes wrong.
Combining everything in a single `if` means that even if the error message gets logged, it still needs to be decoded to understand why something went wrong (which exact requirement was not met).

.cloudbuild/prombench.yaml
saketjajoo: > Maybe a global variable called exit that you can check between steps.

Maybe this might do the trick along with having [allowexitcodes](https://cloud.google.com/build/docs/build-config-file-schema#allowexitcodes). This way the steps can be skipped all the way to the end of the build.

.cloudbuild/prombench.yaml
saketjajoo: Although this is used only once in this step, a separate function was created to have it consistent with line #185. This way, if we need to change the `post_to_github` function, we need not read everything and just change the 2 same functions.

.cloudbuild/prombench.yaml
saketjajoo: The previous gcloud commands get the authentication credentials from GCE metadata server. However, we need to have credentials stored locally for prombench (GCP APIs called in prombench after this step). The reason that this is not before the first gcloud command is that we only need this file if all the previous steps succeed. Otherwise, we need not store the credentials locally if any step fails (which fails the entire build).

.cloudbuild/prombench.yaml
saketjajoo: Yes, I agree. But breaking it into steps means that the environment variables that got changed in previous steps are reset and have the default (stale) values. If so, we'll need to have the `get_or_set_env_var` function in each step to get the updated value.

.cloudbuild/prombench.yaml
saketjajoo: > can create each file Setting environment variables step to avoid the check on line 204

This will make the whole thing consistent, which is good! But, then if later on the environment variables are added or deleted, the file creation step/function would need to be updated as well. Contributors might forget to update the file creation step/function as there are only 3 variables that get updated (as of now). Rest all the env vars are used as is.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
...
+  - 'LAST_COMMIT_SHA='
+  - 'OAUTH_TOKEN='
+  - 'PROMBENCH_IMAGE=$_PROMBENCH_IMAGE'
+  - 'PROMETHEUS_IMAGE_REPOSITORY=gke.gcr.io/prometheus-engine/prometheus'
+  - 'PROVIDER=gke'
+  - 'SERVICEACCOUNT_CLIENT_EMAIL=$_SERVICEACCOUNT_CLIENT_EMAIL'
+  - 'WH_SECRET='
+  - 'ZONE=us-east4-a'
+  substitutionOption: ALLOW_LOOSE
+  dynamicSubstitutions: true
```
maxamins: maybe add a comment explaining reasoning behind flags.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,335 @@
+# https://cloud.google.com/build/docs/build-config-file-schema#options
+options:
```
saketjajoo, line 2: Will move this section after steps.

.cloudbuild/prombench.yaml
maxamins: ignore _PR_NUMBER only available for PR triggers

maxamins: also ignor github repo. Only project_id can be subbed

.cloudbuild/prombench.yaml
maxamins: OK, makes sense. Also names are generic enough.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,490 @@
...
+  - id: "Setting environment variables"
+    name: gcr.io/cloud-builders/gcloud
+    env:
+    - '_PR_COMMENT_TEXT=$_PR_COMMENT_TEXT'
+    allowExitCodes: [1]
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        if [ -f "/workspace/EXIT_EARLY" ]
```
maxamins: Do we need this check for every global variable?

 What if we have a step to create files with default values. That way we know the file exists for sure.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,490 @@
...
+          then
+            echo "Either the given tag is not valid or it is empty. The latest collector tag would be used."
+            DEFAULT_COLLECTOR_TAG=$(gcloud container images list-tags "$${PROMETHEUS_IMAGE_REPOSITORY}" --sort-by=UPDATE_TIME,TAGS | tail -1 | awk '{ print $2 }')
+            echo "$${DEFAULT_COLLECTOR_TAG}" > /workspace/PROMETHEUS_IMAGE_VERSION
+            export PROMETHEUS_IMAGE_VERSION="$${DEFAULT_COLLECTOR_TAG}"
+          else
+            echo "$${COLLECTOR_TAG_OR_CANCEL}" > /workspace/PROMETHEUS_IMAGE_VERSION
+            export PROMETHEUS_IMAGE_VERSION="$${COLLECTOR_TAG_OR_CANCEL}"
+          fi
+
```
maxamins, line 117: 101 and 104 can be on line 107 as echo "$${PROMETHEUS_IMAGE_VERSION}" > /workspace/PROMETHEUS_IMAGE_VERSION

maxamins (commented):
Looking very good! reviewed half, will reviewed half later.

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,495 @@
...
+          then
+            exit 1
+          fi
+        fi
+
+        PROMETHEUS_IMAGE_VERSION=$(cat /workspace/PROMETHEUS_IMAGE_VERSION)
+
+        GITHUB_ORG=$$(echo "${_GIT_REPO}" | cut -d'/' -f1) # GoogleCloudPlatform
+        GITHUB_REPO=$$(echo "${_GIT_REPO}" | cut -d'/' -f2) # prometheus
+        export GITHUB_ORG="$${GITHUB_ORG}" && export GITHUB_REPO="$${GITHUB_REPO}"
```
maxamins: do you actually need export?

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,495 @@
...
+  - 'GRAFANA_ADMIN_PASSWORD=admin'
+  - 'LAST_COMMIT_SHA='
+  - 'OAUTH_TOKEN='
+  - 'PROMBENCH_IMAGE=$_PROMBENCH_IMAGE'
+  - 'PROMETHEUS_IMAGE_REPOSITORY=gke.gcr.io/prometheus-engine/prometheus'
+  - 'PROVIDER=gke'
+  - 'SERVICEACCOUNT_CLIENT_EMAIL=$_SERVICEACCOUNT_CLIENT_EMAIL'
+  - 'WH_SECRET='
+  - 'ZONE=us-east4-a'
+  substitutionOption: ALLOW_LOOSE
```
maxamins: add comment explaining it

maxamins (commented):
Thanks for making the many changes 👍 ! finished reviewing!

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,487 @@
+steps:
+  - id: "Verifying GitHub comment and user permissions"
+    name: bash
+    env:
+    - '_COMMENTER=$_COMMENTER'
```
bwplotka, line 42: Is `_` prefix is some pattern? It's kind of unreadable and odd.  If there's reason, do you mind adding comment why?

.cloudbuild/prombench.yaml
```diff
@@ -0,0 +1,487 @@
...
+    - '_COMMENTER_ROLE=$_COMMENTER_ROLE'
+    - '_GIT_REPO=$_GIT_REPO'
+    - '_PR_COMMENT_ACTION=$_PR_COMMENT_ACTION'
+    - '_PR_COMMENT_TEXT=$_PR_COMMENT_TEXT'
+    - '_PR_NUMBER=$_PR_NUMBER'
+    secretEnv: ['GH_PAT']
```
bwplotka, line 48: What's PAT?

bwplotka (approved):
Thanks! Added a few readability comments/questions but overall LGTM.

Looks more complex and we again put something unique to our fork which will be pain to maintainer, but if there's no other way 🤷🏽

.cloudbuild/prombench.yaml
TheSpiritXIII: Personal Access Token?

saketjajoo: It's GitHub bot's Personal Access Token (to authenticate to GitHub).

.cloudbuild/prombench.yaml
saketjajoo: One of the [rules for declaring user-defined substitution variables](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions) is that they must start with `_`.

PR #115, opened 2023-10-19
scrape: make the initial scrape offset configurable

ridwanmsharif:
This change changes the `NoJitter` boolean configuration to a confiurable `InitialScrapeOffset` option. This will let serverless setups configure more freely how it will want to scrape workloads.

This was needed because in some deployments in serverless environments the scraper (colocated with the workload) might not want to wait an entire scrape interval (since the target could be shortlived) but also might not want to scrape immediately when the target is not ready.

Making this configurable lets the scrape controller (OpenTelemetry Contrib's prometheusreceiver) choose that is appropriate for it depending on its environment.

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -141,12 +141,18 @@ type Options struct {
 	// Optional HTTP client options to use when scraping.
 	HTTPClientOptions []config_util.HTTPClientOption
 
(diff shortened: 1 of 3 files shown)


bwplotka (approved):
Why not 👍🏽  Thanks!

PR #118, opened 2023-11-17
scrape: add option to Manager to skip initial discovery reload

ridwanmsharif:
This change adds a config option to the scrape manager that is quite useful on serverless environments where we are sensitive to the start up latencies of the scraper. The serverless instance might only last for a few seconds and may not be able to afford the minimum 5s reload interval before the scrape pools are created.

This change also adds an option in the discovery manager that will allow users to skip the initial
wait before sending target sets to the scrape manager.

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -134,6 +134,16 @@ func Name(n string) func(*Manager) {
 	}
 }
 
+// SkipInitialWait sets the name of the manager. This is used in serverless flavours of OTel's prometheusreceiver
+// which is sensitive to startup latencies.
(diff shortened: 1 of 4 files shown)


ridwanmsharif:
Rebased on top of https://github.com/GoogleCloudPlatform/prometheus/pull/119 which fixes the flakey test.

bwplotka:
Looks like https://github.com/GoogleCloudPlatform/prometheus/pull/122 is already better, and there is even https://github.com/GoogleCloudPlatform/prometheus/pull/123 so closing this one 🤗

PR #122, opened 2023-11-21
scrape: use locks when shutting down the loops

ridwanmsharif:
Is rebased on top of https://github.com/GoogleCloudPlatform/prometheus/pull/118

This change uses locks in the shutdown path

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -134,6 +134,16 @@ func Name(n string) func(*Manager) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


bwplotka:
This might work, but proposed some small update after research, WDYT? https://github.com/GoogleCloudPlatform/prometheus/pull/123

bwplotka:
OK, cleaned up version is not fully working on discovery without wait feature, so let's merge this one in as it's similar semantically (this PR should work too). We can then iterate on cleanup and better way to avoid initial wait on discovery later.

PR #125, opened 2023-11-22
Prometheus 2.43 GMP support

TheSpiritXIII:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/562

PR #131, opened 2024-01-18
Prometheus 2.45.2 GMP support

TheSpiritXIII:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/772


bwplotka:
Thanks! 💪🏽 Looks good so far

pintohutch:
How do we want to prioritize this against https://github.com/GoogleCloudPlatform/prometheus/pull/125?

TheSpiritXIII:
> How do we want to prioritize this against #125?

Reasons against:
- 2.43 has had more testing and is ready to go. 2.45 has not.
- 2.43 is a smaller step from 2.41 so if we find customer issues, finding the issue and moving back to 2.41 is easier.
- It would delay some projects, e.g. PR #130 is non-trivial to merge into 2.45, mostly due to `prometheus/common` looking different.

bernot-dev:
But on the pros side, 2.45 is an LTS version, and we could stick with it for a while. 🙂

pintohutch:
Then let's focus on merging #125 and getting that into managed collection. We can pursue this separately and release it stand alone.

TheSpiritXIII:
Closing in favor of https://github.com/GoogleCloudPlatform/prometheus/pull/133

PR #133, opened 2024-02-01
Prometheus 2.45.3 GMP support

TheSpiritXIII:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/772


.gcloudignore
```diff
@@ -0,0 +1,3 @@
+.git
```
pintohutch, line 1: nit (non-blocking): What's this file for?

.github/workflows/ci.yml
```diff
@@ -61,25 +61,25 @@ jobs:
...
-    runs-on: ubuntu-latest
-    # The go verson in this image should be N-1 wrt test_go.
-    container:
-      image: quay.io/prometheus/golang-builder:1.20-base
-    steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - run: make build
-      - run: go test ./tsdb/...
-      - run: go test ./tsdb/ -test.tsdb-isolation=false
+  # test_golang_oldest:
```
pintohutch: question: I imagine something breaks with this?

Is there a reason why we need it commented out now, as opposed to before?

.gitignore
```diff
@@ -24,10 +24,9 @@ benchmark.txt
 npm_licenses.tar.bz2
 /web/ui/static/react
 
-/vendor
 /.build
 
-/**/node_modules
+/**/node_modules/.cache
```
pintohutch, line 29: question: I see this was checked in with f3c5df45bbfb88120b2b249f61c75125e10ec4ae, why did we need to change the file at all?

pintohutch: suggestion: could you add a comment explaining?

go.mod
```diff
@@ -202,6 +202,8 @@ require (
 )
 
 replace (
+	// This version is a hard-requirement for content negotiation.
+	github.com/prometheus/common => github.com/TheSpiritXIII/prometheus-common v0.45.0-gmp.0
```
pintohutch, line 209: question: Is this change necessary for the 2.45 fork?

IIUC we're not supporting secrets management with this PR, correct?

pintohutch (commented):
Mostly nits and questions - thanks for putting this together!

go.mod
TheSpiritXIII: I updated the comment to be clearer. Let me know if it doesn't make sense!

.gcloudignore
TheSpiritXIII: To prevent uploading the entire Git repository to Google Cloud Build.  This line is standard in a .gcloudignore file because when you're building, you don't care about the git history. Our file matches the official example: https://cloud.google.com/sdk/gcloud/reference/topic/gcloudignore

I believe that prom-bench would be utilizing this. However, the original change predates our use of prom-bench: https://github.com/GoogleCloudPlatform/prometheus/commit/623141cf6e2b9b6aad2e1a52ad1f6cb7152c9e9c

.github/workflows/ci.yml
TheSpiritXIII: Prometheus' CI builds with two Go versions: the latest and the one specified in the `go.mod`. At the time of creating this PR, prometheus-engine was using 1.21, which meant this fork needed to use the same version. However, since 1.21 was the latest and the oldest we could use, I commented this line out.

Now that 1.22 is available, I've updated this repository. Please take another look. Thanks!

.gitignore
TheSpiritXIII: The same way `go mod vendor` downloads and stores all files in `vendor`, NodeJS downloads and stores all files in `node_modules`. Since we want to contain all vendoring within this repository, we allow `node_modules` as we do `vendor`.  We don't need any more than that to build, so we exclude the `.cache` directory. Otherwise, this directory would grow as we continue doing NodeJS and dependency upgrades.

Does the commit text not already explain the intent?

This change was originally part of the first-ever fork commit: https://github.com/GoogleCloudPlatform/prometheus/commit/623141cf6e2b9b6aad2e1a52ad1f6cb7152c9e9c

.gcloudignore
pintohutch: Oh interesting. Ok - the git blame was confusing here. I'm assuming we technically don't need that anymore. It was probably added for doing manual cloud builds. But happy to keep for now, given precedent.

.github/workflows/ci.yml
pintohutch: SGTM

.gitignore
pintohutch: Ah makes sense. Thanks for clarifying!

go.mod
pintohutch: Discussed offline. tl;dr - this is already folded in our 2.43 fork and enables unit tests to pass.

pintohutch (approved):
LGTM!

Can we do a `/prombench` run?

TheSpiritXIII:
Seeing about at most a 15% reduction in memory with high samples exported. No noticeable CPU usage change.

bwplotka:
I see 10% on collector setup with our export on. Still, great progress!

image

PR #135, opened 2024-02-21
feat: add a native Kubernetes secret provider (v2.45)

TheSpiritXIII:
prometheus-engine change here: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/864

The HTTP Client configuration was expanded to include:

```yaml
basic_auth:
  [ password_ref: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string) ]
authorization:
  [ credentials_ref: [<secret>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#secret) ]
oauth2:
  [ client_secret_ref: [<secret>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#secret) ]
tls_config:
  [ ca_ref: [<secret>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#secret) ]
  [ cert_ref: [<secret>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#secret) ]
  [ key_ref: [<secret>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#secret) ]
```

The [root configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config) now includes:

```yaml
kubernetes_sp_config:
  # The information to access the Kubernetes API.

  # The API server addresses. If left empty, Prometheus is assumed to run inside
  # of the cluster and will discover API servers automatically and use the pod's
  # CA certificate and bearer token file at /var/run/secrets/kubernetes.io/serviceaccount/.
  [ api_server: [<host>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#host) ]

  # Optional path to a kubeconfig file.
  # Note that api_server and kube_config are mutually exclusive.
  [ kubeconfig_file: [<filename>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#filename) ]

kubernetes_secrets:
# The name of the secret (used in secret ref in HTTPClientConfig.
- name: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string)
  config:
    # The namespace of the Kubernetes Secret to use.
    namespace: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string)
    # The name of the Kubernetes Secret to use.
    name: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string)
    # The key of the Kubernetes Secret to use.
    key: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string)
```

This is all protected under the `google-kubernetes-secret-provider` feature flag, to be replaced should we merge this functionality upstream.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	gcm_export "github.com/GoogleCloudPlatform/prometheus-engine/pkg/export/setup"
+	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/secrets"
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/alecthomas/units"
 	"github.com/go-kit/log"
@@ -43,6 +44,7 @@ import (
 	"github.com/mwitkow/go-conntrack"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
+	common_config "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/promlog"
 	promlogflag "github.com/prometheus/common/promlog/flag"
@@ -154,6 +156,7 @@ type flagConfig struct {
 	enableNewSDManager         bool
 	enablePerStepStats         bool
 	enableAutoGOMAXPROCS       bool
+	enableKubeSecretProvider   bool
 
 	prometheusURL   string
 	corsRegexString string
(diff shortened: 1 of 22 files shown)


cmd/prometheus/main.go
```diff
@@ -627,6 +634,21 @@ func main() {
...
+			prometheus.DefaultRegisterer,
+			secrets.ProviderOptions{
+				Logger: log.With(logger, "component", "secret manager"),
+			},
+		)
+		secretManager = &manager
+		defer secretManager.Close(prometheus.DefaultRegisterer)
+	}
+
+	cfg.scrape.HTTPClientOptions = append(cfg.scrape.HTTPClientOptions, common_config.WithSecretManager(secretManager))
```
pintohutch: question: In the case where the flag is disabled, I'm assuming it's safe to add a `nil` reference for the uninitialized `secretManager` pointer?

What about just not adding the manager at all in that case?

pintohutch (commented):
Can you add some examples of what fields you're adding in the PR description so we have an idea of how the configuration surface is changing here?

There are some sketches in the [proposal](https://docs.google.com/document/d/1EqHd2EwQxf9SYD8-gl3sgkwaU6A10GhiN7aw-2kx7NU/edit), but it's unclear to me how much of that is implemented here.

cmd/prometheus/main.go
TheSpiritXIII: Good catch. Thanks!

pintohutch (approved):
Thanks for the configuration example in the PR - great reference!

PR #145, opened 2024-03-22
chore: enable cgo and use google buildbase

pintohutch:
cgo needs to be enabled to link against boringcrypto, so we add that here.

In addition, we use the google-go.pkg.dev/golang image as the Go buildbase to ensure build-time requirements, like boringcrypto, are enabled.

We also use gke.gcr.io/gke-distroless/libc as our runtime image.

We add the "cryp/tls/fipsonly" import to ensure boringcrypto is linking properly at build time. We guard this with a build tag "boring" in a dedicated file.

Finally, we remove promtool from the image, as its not intended to be used.

Cherry-pick 6cd5dd0bb63e12eed80f5f87ef775ae399196741

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG IMAGE_BUILD_NODEJS=launcher.gcr.io/google/nodejs
-ARG IMAGE_BUILD_GO=golang:1.20-bullseye
+ARG IMAGE_BUILD_GO=google-go.pkg.dev/golang:1.20.14@sha256:6f86d8a81ff191bee8d3ff8b4c193889560b4ca15df373d5084953c5c860190f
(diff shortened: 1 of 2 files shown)


pintohutch:
Results look good.

/prombench cancel

Dockerfile
```diff
@@ -48,7 +43,6 @@ RUN ["/busybox/sh", "-c", "mkdir -p /prometheus"]
 FROM ${IMAGE_BASE}
 
 COPY --from=buildbase /app/prometheus /bin/prometheus
-COPY --from=buildbase /app/promtool /bin/promtool
```
TheSpiritXIII, line 51: praise: Good catch! I just created a bug about this lol

TheSpiritXIII (approved):

PR #153, opened 2024-04-03
Add deletion-data-on-start flags as per go/gmp:oomcrashloop

bwplotka:
See go/gmp:oomcrashloop and https://github.com/prometheus/prometheus/issues/13939

I decided to slightly alter the flag name and instead create a specific fork GMP flag so it's explicit it's experimental and only in our fork. Some `--storage.agent.delete-data-on-start` might be accepted for agent, but unlikely for TSDB, also names might change, so let's not confuse users.


cmd/prometheus/main_test.go
```diff
@@ -357,6 +357,71 @@ func getCurrentGaugeValuesFor(t *testing.T, reg prometheus.Gatherer, metricNames
 	return res
 }
 
+func TestDeleteStorageDataOnStart(t *testing.T) {
+	for _, agentMode := range []bool{false, true} {
+		t.Run(fmt.Sprintf("%v", agentMode), func(t *testing.T) {
```
pintohutch, line 362: question: Where is prometheus invoked here? How is it running and deleting these dirs?

bwplotka: It's not, I only test the function which in invoked in `main.go`. The invocation is trivial, plus it's very difficult to test this by running Prometheus and validating e.g. old samples (on CI). We need proper e2e framework, at some point we could propose this to Prometheus (e.g. https://github.com/efficientgo/e2e) - I planned that for remote write things anyway.

bwplotka: Once shipped I will manually test things with my setup, just in case.

TheSpiritXIII (approved):
Looks good to me! Great work.

cmd/prometheus/main_test.go
pintohutch: Oh - nvm. I was bleary-eyed when reviewing this and missed the `deleteStorageData` call on L366.

pintohutch (approved):

PR #179, opened 2024-06-05
feat: Prometheus export.go update for dynamic configs

TheSpiritXIII:
Pre-req: https://github.com/GoogleCloudPlatform/prometheus/pull/180


cmd/prometheus/main.go
```diff
@@ -871,7 +881,7 @@ func main() {
 			name: "gcm_export",
 			reloader: func(cfg *config.Config) error {
 				// Call in closure to not call Global() before it's initialized below.
-				return gcm_export.Global().ApplyConfig(cfg)
+				return gcm_export.Global().ApplyConfig(cfg, nil)
```
pintohutch: question: Don't you need to pass in the `exporterOpts` on reload instead of `nil`?

cmd/prometheus/main.go
```diff
@@ -938,8 +948,17 @@ func main() {
 		)
 	}
 	{
+		exporterLogger := log.With(logger, "component", "gcm_exporter")
```
pintohutch, line 947: question: Is there any way to wrap this up in a constructor, i.e. `newExporter` call?

It looks like the client (i.e. Prometheus) has a bit more explicit setup work to do now than before, which is less than ideal in a world where we're trying to defork ourselves. Is it feasible to consolidate these in a constructor in our export package?

A lot of these functions look new to me and I'm not exactly sure how they work, so I may just be out of the loop.

pintohutch (commented):
Just giving my feedback - not blocking. I'll let @bwplotka weigh in.

cmd/prometheus/main.go
TheSpiritXIII: It's optional but I added it. Thanks!

cmd/prometheus/main.go
TheSpiritXIII: Done. Thanks!

go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/prometheus/prometheus
 
-go 1.21
+go 1.22.3
```
pintohutch, line 3: question (non-blocking): Is this change related?

I have nothing against bumping go for the module. But if it's not necessary, doing this in a separate PR or commit will make it easier to isolate these changes as they roll out.

cmd/prometheus/main.go
```diff
@@ -871,7 +877,7 @@ func main() {
 			name: "gcm_export",
 			reloader: func(cfg *config.Config) error {
 				// Call in closure to not call Global() before it's initialized below.
-				return gcm_export.Global().ApplyConfig(cfg)
+				return gcm_export.Global().ApplyConfig(cfg, &opts.ExporterOpts)
```
pintohutch, line 880: question (non-blocking): Ah - IIUC you originally left the arg `nil` as the reload function would never trigger at runtime anyway, since the opts (as of this PR) can only change when the process restarts. Do I have that correct?

pintohutch (approved):
LGTM % 2 non-blocking questions

cmd/prometheus/main.go
TheSpiritXIII: That's correct!

go.mod
TheSpiritXIII: `go mod tidy` updates the Go version to match the smaller of the minimum Go version of dependencies OR the current version in `go.mod`.

In this case, it updated to because prometheus-engine was using this version: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/go.mod

PR #180, opened 2024-06-05
feat: allow updating export flags at runtime

TheSpiritXIII:
This PR allows you to set Google Cloud Monitoring API export-related flags in the configuration, as opposed to the CLI. The main motivation for this change is to eliminate UPDATE RBAC permissions which are otherwise required to update the CLI flags at runtime in a Kubernetes environment.

The following are now settable in the Prometheus config:
```yaml
google_cloud:
  export:
    match: []string
    compression: string
    credentials: string
```

When both CLI arguments and configurations are present, the CLI arguments will be the default and the set configurations will override the CLI arguments.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -69,6 +69,7 @@ import (
 	"github.com/prometheus/prometheus/notifier"
 	_ "github.com/prometheus/prometheus/plugins" // Register plugins.
 	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/scrape"
 	"github.com/prometheus/prometheus/storage"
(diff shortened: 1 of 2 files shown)


cmd/prometheus/main.go
```diff
@@ -880,8 +881,38 @@ func main() {
 		}, {
 			name: "gcm_export",
 			reloader: func(cfg *config.Config) error {
+				// Don't modify defaults. Copy defaults and modify based on config.
```
pintohutch, line 880: question: Do you need both `exporterOpts` and `newExporterOpts` here?

pintohutch (commented):
Thanks - could you fill out the PR description to give a summary of the changes?

cmd/prometheus/main.go
TheSpiritXIII: Good point. It is a little redundant. I got rid of the latter.

pintohutch (approved):
:shipit:

PR #181, opened 2024-06-05
chore: update to Golang 1.22.3 to match prometheus-engine

TheSpiritXIII:
This is necessary to update our prometheus-engine further.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,17 +61,18 @@ jobs:
           go test $TestTargets -vet=off -v
         shell: powershell
 
-  test_golang_oldest:
-    name: Go tests with previous Go version
-    runs-on: ubuntu-latest
-    # The go verson in this image should be N-1 wrt test_go.
-    container:
-      image: quay.io/prometheus/golang-builder:1.21-base
-    steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - run: make build
-      - run: go test ./tsdb/...
-      - run: go test ./tsdb/ -test.tsdb-isolation=false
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -61,17 +61,17 @@ jobs:
...
-    # The go verson in this image should be N-1 wrt test_go.
-    container:
-      image: quay.io/prometheus/golang-builder:1.21-base
-    steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - run: make build
-      - run: go test ./tsdb/...
-      - run: go test ./tsdb/ -test.tsdb-isolation=false
+  # test_golang_oldest:
+  #   name: Go tests with previous Go version
```
maxamins, line 66: why comment this out? maybe leave a comment explaining.

maxamins (approved):

.github/workflows/ci.yml
TheSpiritXIII: We only support Golang 1.22. Golang 1.21 is unsupported and 1.23 does not exist. There is no "oldest" Golang version to test.

maxamins: makes sense. 
```suggestion
  # Commented out because GMP doesn't support previous version of go.
  # test_golang_oldest:
  #   name: Go tests with previous Go version
```

.github/workflows/codeql-analysis.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '>=1.21 <1.22'
+          go-version: '>=1.22 <1.23'
```
bernot-dev: This is equivalent to `~1.22`

PR #189, opened 2024-06-20
make prometheus worse

Adib234:


ridwanmsharif:
Does the prombench benchmark cluster have to be running still?

pintohutch:
Nope - kill it

PR #205, opened 2024-10-05
chore: rebase feature/shutdownscrape on upstream/main

ridwanmsharif:
chore: rebase feature/shutdownscrape on upstream/main
    
    This change does the following:
     - Rebase the feature branch on top of main to get rid of some vulns and
       some dependency issues that popped up when upgrading the cloud run
       sidecar
     - Fix tests by adding more robust and clear instructions on when to
       notify downstream managers of changes to the discovered targets.


ridwanmsharif:
I feel like maybe I don't know what I'm doing????

I need to update the https://github.com/GoogleCloudPlatform/run-gmp-sidecar/ repo with the latest OTel version components and it relies on this branch. OTel also relies on some common prometheus libraries and in the sidecar repo, we replace prometheus libraries with equivalent one from this branch.

But things diverged and lots of things broke when I upgraded. I tried in this PR to rebase `feature/shutdownscrape` on `upstream/main` and it does work. The last few commits are all that really matter in this change, but since its a rebase, the PR looks awful. Is there a better way?

Should I just push to another branch and sunset `feature/shutdownscrape` in favor of that branch?

The last commit does make some changes that need to be reviewed though - was needed to make tests work but I need an adult to sanity check what is happening in there in the discovery and scrape managers.

PR #207, opened 2024-10-07
gmp: Skip WAL (passthrough) if no PRW is specified, normal agent mode otherwise.

bwplotka:
Before this change GMP fork in agent mode for 2.45.3 will write data to agent DB/WAL and rely on truncation to remove the entries. This is wasteful if no PRW is configured as we have a side channel for GCM export.

This change literally skips all sample appends if no PRW is configured. This effectively keeps agent DB/WAL always empty. If new configuration comes in with PRW entry, the append is enabled and all should work as in vanilla Prometheus.

This should give us maximum efficiency, while keeping agent mode smooth for everyone. This also allows experiments with "passthrough" mode (no WAL/persistence mode) in future of Prometheus.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -771,6 +771,7 @@ func main() {
 	// This is passed to ruleManager.Update().
 	externalURL := cfg.web.ExternalURL.String()
 
+	gcmAgentWriteSkipper := &writeSkipperForNoRWConfig{logger: logger, noRWEndpointConfigured: atomic.NewBool(true)}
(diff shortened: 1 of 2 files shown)


bwplotka:
Kind of works, but some statistical bugs are there, to investigate:

image

bwplotka:
As per https://github.com/bwplotka/prombenchy/pull/14: 

CPU is same/similar to srv mode or unchanged agent. Memory is initially much better, but then leaking (for both this PR and unchanged agent, confirming what some cx saw). To be investigated.

image

PR #208, opened 2024-10-08
chore: rebase `feature/shutdownscrape` on v2.54.1

ridwanmsharif:
Cherry-picks the commits from `feature/shutdownscrape` onto a mirror of v0.54.1


braydonk (approved):

bwplotka (approved):
I think you mean 2.54.1 (I checked as well). Maybe 0. is because we don't have v2 in go.mod, confusing the go modules typically.

Anyway, LGTM, looking forward to having this in upstream one day 💪🏽

PR #214, opened 2024-12-09
fix: npm deps

bernot-dev:
Run `npm install` for ui to get tests passing.


bwplotka (approved):
Might just out-dated packages then. Thanks!

PR #217, opened 2024-12-12
chore: use newer distroless image

### LINKED ISSUE #212: Is there any planning on using debian12 base image?
The latest prometheus image seems build from a image with debian 11 image
gke.gcr.io/gke-distroless/libc@sha256:4f834e207f2721977094aeec4c9daee7032c5daec2083c0be97760f4306e4f88
we identify it by looking at the etc/os-release file in the layer.

Also I see the there was a PR switch back on using debian11 image from debian 12 due to image not exist.
And now prometheus is using another set of image that is from gke.gcr.io/gke-distroless/libc which seems the latest image is still in debian11.
( etc/os-release still shows debian11 )

Since debian 11 is end of life now, is there any planning migrating to use a debian12 image?
or if there is any agreement from google on vulnerability patching that it is fine to use the gke-distroless/libc image even it is a with debian11 os?

bernot-dev:
Use `base-nossl` instead of `libc` image as base for Prometheus. `base-nossl` is technically based on Debian 12 instead of Debian 11.

Fixes #212


bwplotka (approved):
Thanks!

PR #219, opened 2025-01-23
Updated x/net and x/crypto to latest.

bwplotka:
+ VERSION bump

Security fix.


bwplotka:
not going to worry about UI tests, we didn't change anything. We need fresh prom version.

go.mod
```diff
@@ -185,10 +185,10 @@ require (
 	go.mongodb.org/mongo-driver v1.13.1 // indirect
 	go.opencensus.io v0.24.0 // indirect
 	go.opentelemetry.io/otel/metric v1.22.0 // indirect
-	golang.org/x/crypto v0.31.0 // indirect
+	golang.org/x/crypto v0.32.0 // indirect
```
pintohutch, line 188: I think 0.31.0 for x/crypto is ok. Did this change based on bumping x/net?

pintohutch (approved):

go.mod
bwplotka: Yes, also I wanted latest of this just as a good practice.

PR #235, opened 2025-07-04
google patch[setup]: fix build and clean readme.

bwplotka:
Essentially making sure we have any test/build for an early warning before it hits our internal CI


bwplotka:
Assuming GH action is OOMing; keeping like this for now. 🤔

PR #246, opened 2025-08-06
feat: Add populate description from help support

bwplotka:
cc @lyanco @realschwa 

It looks as it works 🎉 

image

~I don't see our [Metadata API to give it back though](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata)... but that's ok for now? I can switch to using List time series to test this.~ EDIT: ignore me, it works, my test was wrong.

Q: Do we need feature flag for this? I literally push it hard with frequent changes and it picks the first one only as expected, looks like we can rely on this well. I think I will revert the flag addition in this PR, I propose we switch without flag for this WDYT? 🤔 @pintohutch @bernot-dev ?

### DIFF
--- a/google/export/series_cache.go
+++ b/google/export/series_cache.go
@@ -64,6 +64,14 @@ type seriesCache struct {
 
 	// Prefix under which metrics are written to GCM.
 	metricTypePrefix string
+
+	// Feature flag that controls passing the HELP of the metric as a GCM
+	// metric description. This was historically not done to avoid the GCM
+	// limitation in handling description changes, often possible in OSS metric
+	// world (Prometheus, Otel). As of 2025 Q1 this has been now fixed, but
+	// we maintain feature flag so we can rollback on scale in case of the
+	// handling issues.
+	populateDescription bool
 }
 
 type seriesCacheEntry struct {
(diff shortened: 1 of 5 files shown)


pintohutch:
Awesome!

> Do we need feature flag for this?

IMO this is an enhancement we can capture in a version bump and our release notes without a flag. Not having description available in the metric itself was a gap in existing Prometheus functionality, no?

bwplotka:
> IMO this is an enhancement we can capture in a version bump and our release notes without a flag. Not having description available in the metric itself was a gap in existing Prometheus functionality, no?

Agree, the only reason for flag would be an incident risk due to bad handling of rapid help changes on GCM side, so wanted to confirm the risk is low here (:

google/export/series_cache.go
```diff
@@ -133,10 +133,10 @@ func (e *seriesCacheEntry) setNextRefresh() {
 }
 
 func newSeriesCache(
-	logger log.Logger,
-	reg prometheus.Registerer,
-	metricTypePrefix string,
-	matchers Matchers,
+		logger log.Logger,
```
bwplotka: my IDE formatting hates me

bernot-dev (approved):
LGTM

PR #249, opened 2025-08-07
feat: configurable lease durations

damdinsharav:
This change adds three new optional command-line flags and corresponding environment variable overrides to configure HA lease timing parameters:

--export.ha.lease-duration (env: LEASE_DURATION)

--export.ha.renew-deadline (env: RENEW_DEADLINE)

--export.ha.retry-period (env: RETRY_PERIOD)

These flags are now available in HAOptions.SetupFlags without explicit default values, leaving defaulting responsibility to the lease package’s runtime logic.

This approach:

Enables users to override lease timing behavior explicitly via CLI or env vars

Keeps default lease durations and retries centralized and consistent in the lease implementation

No functional changes to lease behavior have been made. Only configuration is exposed for easier tuning.

### DIFF
--- a/google/export/setup/setup.go
+++ b/google/export/setup/setup.go
@@ -204,6 +204,10 @@ type HAOptions struct {
 	KubeConfigFile string
 	KubeNamespace  string
 	KubeName       string
+
+	LeaseDuration time.Duration
+	RenewDeadline time.Duration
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
Thanks, LGTM!

PR #237, opened 2025-07-10
feat: cut release 2.53.5-gmp

bwplotka:
This is a first release that uses:

* New module architecture (go/gmp:fork-toil) where:
  * this code does NOT import prometheus-engine
  * prometheus-engine imports forked code
  * export,secret,lease are part of the forked code (see 2nd commit)
* Scripted creation (go/gmp:toil-automation)

```
SOURCE_BRANCH=release-2.45.3-gmp UPSTREAM_TAG=v2.53.5 CHECKOUT_DIR=~/tmp-release bash hack/release-forksync.sh
```

Also used some pieces and learnings from @pintohutch  https://github.com/GoogleCloudPlatform/prometheus/tree/release-2.53.4-gmp

PR #251, opened 2025-08-12
feat: Changed node image; Renamed to Dockerfile.google; Add vendor target to avoid internal CI node version issues.

bwplotka:
Rename: 
* Follow up after https://github.com/GoogleCloudPlatform/prometheus/pull/237#discussion_r2236716386

NodeJS image change:
* the current nodejs image is archived. Replaced it with the well maintained (https://cloud.google.com/docs/buildpacks/base-images) `us-central1-docker.pkg.dev/serverless-runtimes/google-22/runtimes/nodejs22`. (Thanks @bernot-dev for finding this!).  We do need new image as for npm install, current Prom versions require newer node/npm, whereas building work with the old version.

Vendor target:
* I tried generic go mod and npm vendoring script on Louhi side, but eventually it fails due to npm version mismatches. We can install correct version, but likely, there's a risk that no "correct" version that will work against all versions of forks exists for all cases.
* Much better and giving quicker feedback way is to assume that our repos have google_vendor.sh script that can be invoked on sync and import flows. We can also use the same vendoring script in local CI to ensure we didn't break the release flow.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG IMAGE_BUILD_NODEJS=launcher.gcr.io/google/nodejs
+ARG IMAGE_BUILD_NODEJS=node:current@sha256:e8e882c692a08878d55ec8ff6c5a4a71b3edca25eda0af4406e2a160d8a93cf2
```
bwplotka: We might need to get some help on this one internally or use Prometheus/official node image here.. We can't use launcher.gcr.io/google/nodejs anymore.

bwplotka: @bernot-dev found us-central1-docker.pkg.dev/serverless-runtimes/google-22/runtimes/nodejs22 which seems to well maintained and fits our needs!

bwplotka:
Update: Vendoring locally installs `@lazer/generator`, but in docker, it somehow skips it -- should be last thing to fix! On it from Monday.

image

bwplotka:
Figured this out, should be good to go; PTAL @bernot-dev @pintohutch  (pinged you internally about corresponding CI change)

Dockerfile.google
```diff
@@ -0,0 +1,115 @@
+# This dockerfile is multi target. Use DOCKER_BUILDKIT=1 when building and reference the target:
+# --target=vendor with -o <host local path> to build image with the generated dependencies to copy and vendor on demand.
+# --target=app to build actual application image.
+
+# For the lack of the other official Google nodejs image, we use serverless project
+# images to build the Prometheus frontent (https://cloud.google.com/docs/buildpacks/base-images).
+#ARG IMAGE_BUILD_NODEJS=us-central1-docker.pkg.dev/serverless-runtimes/google-22/runtimes/nodejs20:latest@sha256:25c59b8fa439beb8cc9047677eb3edcad8df614d594579058b0c15176e53a6b4
```
bwplotka: Remove this.

bernot-dev (approved):
LGTM

PR #253, opened 2025-08-18
feat: Add prw2gcm binary

bwplotka:
Relates to b/425614387

This adds prw2gcm binary to our Prometheus fork image that forwards PRW 2.x requests as GCM v3 gRPC calls.

NOTE: nhcb and potential classic histogram support will be added in the next PRs.

## Rationales

While initially we dismissed writing this proxy, given the potential OpenTelemetry Collector's PRW2.0 -> Otel data model -> GCM/OTLP flow recent (experimental) development... I decided to write it anyway 🙈  I put some limited time for it, but I wanted to answer some questions:
* How hard is to implement conversion with the improvements we made in PRW 2.x?
* How usable is [our new experimental Go library for PRW](https://github.com/prometheus/client_golang/tree/main/exp#remote).
* How the handling looks like if we [change PRW 2.x  CT proto structure slightly](https://github.com/prometheus/prometheus/pull/17036)

Turns our the conversion is straightforward for Prometheus use (single sample) and a little bit more involved with multiple samples, but still... manageable to write (with some e2e tests!) in ~8 SWE-h 🎉  Also spent some extra time trying to figure out how to pass freaking auth token from the HTTP header into gRPC  call 🙈

Apart from learnings, I do think it's useful to maintain and even ship with our Prometheus fork image, because:

* Our GCM ingestion can be inspired (or even straight import) this code.
* GCM PRW support is incoming, but will take time. With this prw2gcm binary we can:
  * Start heavy testing of vanilla Prometheus -> GCM, without significant complexities of the untested Otel path. I plan to use this binary in our [e2e GCM tests](https://github.com/GoogleCloudPlatform/prometheus/tree/release-2.53.5-gmp/google/internal/promqle2etest) in the next PRs
 
EDIT: See https://github.com/GoogleCloudPlatform/prometheus/pull/255 for tests using prw2gcm.  
  
  * Users will be able to use vanilla Prometheus with prw2gcm sidecar from day 1, allowing us to unfork practically sooner. Again, without Otel 3 model conversions complexity and overhead.


google/cmd/prw2gcm/proxy.go
```diff
@@ -0,0 +1,300 @@
...
+		queueCh:       make(chan *monitoring_pb.TimeSeries, 10),
+		gcmCreateTSFn: gcmCreateTSFn,
+	}
+	q.wg.Add(1)
+
+	go q.run()
+	return q
+}
+
+const maxBatchSize = 100
```
maxamins: Should be 200?

pintohutch:
Thanks Bartek - what's your rationale specifically for not creating a dedicated `prw2gcm` image?

google/cmd/prw2gcm/proxy.go
```diff
@@ -0,0 +1,300 @@
...
+			return w, err
+		}
+	}
+	projectID := getProjectID(req.Context())
+	if projectID == "unknown" {
+		// Programmatic error, should not happen if the detectPOSTMethodProjectID middleware is set.
+		return nil, newHTTPError(http.StatusInternalServerError, "no project id found")
+	}
+
+	serializedRequest, err := io.ReadAll(req.Body)
```
maxamins, line 186: what about snappy compression?

google/cmd/prw2gcm/convert.go
```diff
@@ -0,0 +1,397 @@
...
+			extraSuffix = gcmMetricSuffixNone
+			kind = metric_pb.MetricDescriptor_GAUGE
+		default:
+			return "", kind, newHTTPErrorf(http.StatusBadRequest, "unknown summary series suffix %v", ms)
+		}
+	case writev2.Metadata_METRIC_TYPE_INFO, writev2.Metadata_METRIC_TYPE_STATESET:
+		// TODO: Check if it's safe to do (new behaviour vs export pkg).
+		suffix = gcmMetricSuffixGauge
+		kind = metric_pb.MetricDescriptor_GAUGE
+	case writev2.Metadata_METRIC_TYPE_UNSPECIFIED:
```
maxamins, line 347: should we use the [unknown suffix here? ](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/d63201438542c3fbc7685c6f2175d0aae5b2f33d/pkg/export/series_cache.go#L420)can be in a future cl.

google/cmd/prw2gcm/convert.go
```diff
@@ -0,0 +1,397 @@
...
+			continue
+		}
+		if maxPoints < points {
+			maxPoints = points
+		}
+		initialSeries = append(initialSeries, exportTS)
+		initialSeriesTsIndex = append(initialSeriesTsIndex, i)
+	}
+
+	// Go through samples. We have to do it over sample dimension, not series, given the
```
maxamins, line 86: Can we just iterate over the series dimension and send each point in it's own CreateTimeSeries request?

maxamins (approved):

google/cmd/prw2gcm/proxy.go
bwplotka: Good question! It's built into our handy PRW library (https://github.com/prometheus/client_golang/blob/84a4734fe73aecc9ec36e1fbcb09c9a2d679f392/exp/api/remote/remote_api.go#L460)

google/cmd/prw2gcm/convert.go
bwplotka: Would it be efficient enough for everyone to have batchLen = 1 always?

bwplotka:
> Thanks Bartek - what's your rationale specifically for not creating a dedicated prw2gcm image?

Just simplicity, similarly we don't have separate image for `promtool`. It is meant to be a tmp thing next to Prometheus, so feels like a good enough choice.

We can make it separate, no problem. Should we? If yes, should it be in prometheus-engine that already builds multiple images or here? 

I like having it here, because:
*  We can test directly in our e2e gcm test suite which is maintained (and run on CI here). In fact we test *exact* version of the proxy binary since the code for it is here too.
* This proxy depends on Prometheus AND export package, so putting it to operator module that has more deps feels unnecessary.

google/cmd/prw2gcm/proxy.go
bwplotka: Ups, good catch, I don't know why I put 100 here, I did that in the draft of this work like 1y ago...

I see in export it's 200 and it's configuration (https://github.com/GoogleCloudPlatform/prometheus/blob/ba13c6f51cee94d16c3e91da2e38dd00959a73bb/google/export/export.go#L162), so let's change to 200 here now!

google/cmd/prw2gcm/convert.go
bwplotka: Added TODO for this, but I wonder if this will be necessary - at the receiving point it won't be really useful to ingest unknown metrics (no CT likely)

pintohutch:
Those reasons seem compelling enough to me to keep it in this image :)

bwplotka:
Let's merge the initial state - we won't be mentioning this tool anywhere until it's ready.

Thanks!

PR #258, opened 2025-08-29
fix: use boringcrypto

bernot-dev:
Use BoringCrypto


bwplotka (approved):
Thanks!

bwplotka:
Test failure is unrelated - some old test cert got expired recently. We can fix later.

PR #259, opened 2025-09-12
feat(2.53.5): opt-in fix for export.match from config; ensure ApplyConfig before export

bwplotka:
Implements step 1 from go/gmp:matchstuck

NOTE: This PR also expects `ApplyConfig` to be triggered for export to work. Otherwise it's noop (similar to disable export = true). IMO this is required for reliable startup behaviour for all options (especially matching). Imagine flag is setting = matchAll and config is setting match only "A". We don't want risk random metrics user don't expect (e.g. broken ones) due to startup races. 

In practice Prometheus ensures ApplyConfig is triggered before DB is ready, but adding safecheck for other callers.

### DIFF
--- a/google/config/config.go
+++ b/google/config/config.go
@@ -2,6 +2,9 @@ package config
 
 // GoogleCloudConfig represents a custom Prometheus config entry
 // required by the hot-reloading requirements in high security k8s setups.
+//
+// All configuration parameters overrides the corresponding EXTRA_ARGS environment
+// or CLI flags.
 type GoogleCloudConfig struct {
 	Export GoogleCloudExportConfig `yaml:"export,omitempty"`
 	Query  GoogleCloudQueryConfig  `yaml:"query,omitempty"`
@@ -11,10 +14,24 @@ type GoogleCloudConfig struct {
 type GoogleCloudExportConfig struct {
 	// Compression controls the ExporterOpts.Compression setting.
 	Compression string `yaml:"compression,omitempty"`
-	// CredentialsFile controls ExporterOpts.CredentialsFile
+	// CredentialsFile controls the ExporterOpts.CredentialsFile setting.
 	CredentialsFile string `yaml:"credentials,omitempty"`
 
-	// Deprecated: This option no longer works, see https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1688
+	// EnableMatch allows additional control over matching filtering
+	// given the go/gmp:matchstuck context.
+	//
+	// Available settings:
+	// * not set: This config's Match settings are ignored. Custom export.match flags are used.
+	// * false: Filtering feature is explicitly disabled; export matches all series.
+	// * true: This config's Match settings are used. Overwrites all custom export.match flags.
+	EnableMatch *bool `yaml:"enable_match,omitempty"`
+	// Match, if EnableMatch is true, controls the ExporterOpts.Matchers setting.
+	// It offers runtime change-able control of the "matchOneOf" filtering.
+	// This filtering will skip certain series from entering the series cache (for export).
+	// This data will still be ingested into Prometheus local storage.
+	//
+	// IMPORTANT: This option is prone to misconfiguration, thus opt-in and removed from the public docs.
+	// See https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#filter-metrics.
 	Match []string `yaml:"match,omitempty"`
 }
 
--- a/google/export/export.go
+++ b/google/export/export.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -149,6 +150,13 @@ type Exporter struct {
 	// Used to construct a new metric client when options change, or at initialization. It
 	// is exposed as a variable so that unit tests may change the constructor.
 	newMetricClient func(ctx context.Context, opts ExporterOpts) (metricServiceClient, error)
+
+	// Currently applied matchers in seriesCache.matchers. Avoids locking for reads.
+	matchers Matchers
+
+	// We allow runtime config path. For startup consistency make exporter noop until export
+	// will see the first ApplyConfig.
+	configured bool
 }
 
 const (
@@ -206,7 +214,8 @@ type ExporterOpts struct {
 	// A list of metric matchers. Only Prometheus time series satisfying at
 	// least one of the matchers are exported.
 	// This option matches the semantics of the Prometheus federation match[]
-	// parameter.
+	// parameter. This is a setting from flags only, see ApplyConfig on
+	// how it's applied dynamically in conjunction with Export.Match runtime option.
 	Matchers Matchers
 
 	// Prefix under which metrics are written to GCM.
@@ -367,6 +376,7 @@ func defaultNewMetricClient(ctx context.Context, opts ExporterOpts) (metricServi
 }
 
 // New returns a new Cloud Monitoring Exporter.
+// IMPORTANT: Created exporter is noop until ApplyConfig is called at least once. This is for configuration startup consistency.
 func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts ExporterOpts, lease Lease) (*Exporter, error) {
 	grpc_prometheus.EnableClientHandlingTimeHistogram(
 		grpc_prometheus.WithHistogramBuckets([]float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 20, 30, 40, 50, 60}),
@@ -409,7 +419,7 @@ func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts
 		ctx:                  ctx,
 		opts:                 opts,
 		metricClient:         metricClient,
-		seriesCache:          newSeriesCache(logger, reg, opts.MetricTypePrefix, opts.Matchers),
+		seriesCache:          newSeriesCache(logger, reg, opts.MetricTypePrefix),
 		externalLabels:       createLabelSet(&config.Config{}, &opts),
 		newMetricClient:      defaultNewMetricClient,
 		nextc:                make(chan struct{}, 1),
@@ -418,6 +428,10 @@ func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts
 		lease:                lease,
 	}
 
+	// Set initial matchers.
+	e.matchers = opts.Matchers
+	e.seriesCache.setMatchers(e.matchers)
+
 	// Whenever the lease is lost, clear the series cache so we don't start off of out-of-range
 	// reset timestamps when we gain the lease again.
 	lease.OnLeaderChange(e.seriesCache.clear)
@@ -425,7 +439,6 @@ func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts
 	for i := range e.shards {
 		e.shards[i] = newShard(opts.Efficiency.ShardBufferSize)
 	}
-
 	return e, nil
 }
 
@@ -442,10 +455,12 @@ const (
 // ApplyConfig updates the exporter state to the given configuration. For the
 // Prometheus fork config's google_cloud entry, ExporterOpts might be
 // changed and applied to the exporter, potentially recreating the metric client.
+// NOTE: Runtime configuration will not override disable/disableAuth options from flags.
 func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
 	// Note: We don't expect the NopExporter to call this. Only the config reloader calls it.
 	e.mtx.Lock()
 	defer e.mtx.Unlock()
+	e.configured = true
 
 	// Don't recreate the metric client each time. If the metric client is recreated, it has to
 	// potentially redo the TCP handshake. With HTTP/2, TCP connections are kept alive for a small
@@ -463,7 +478,6 @@ func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
 		e.opts.CredentialsFile = cfg.GoogleCloud.Export.CredentialsFile
 		recreateClient = true
 	}
-	// NOTE(bwplotka): Matchers are deprecated, see https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1688
 
 	lset := createLabelSet(cfg, &e.opts)
 	labelsChanged := !labels.Equal(e.externalLabels, lset)
@@ -477,20 +491,41 @@ func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
 		}
 	}
 
-	// If changed, or we're calling this for the first time, we need to recreate the client.
+	var toApply Matchers
+	switch {
+	case cfg.GoogleCloud.Export.EnableMatch == nil:
+		// Ignore Export.Match, whatever was in flag or default wins.
+		toApply = e.opts.Matchers
+	case !*cfg.GoogleCloud.Export.EnableMatch:
+		// Explicit apply all matching (disabling the filtering).
+		toApply = nil
+	case *cfg.GoogleCloud.Export.EnableMatch:
+		// Runtime Export.Match opted-in; override:
+		for _, match := range cfg.GoogleCloud.Export.Match {
+			if err := toApply.Set(match); err != nil {
+				return err
+			}
+		}
+	}
+
 	if recreateClient {
+		// If changed, or we're calling this for the first time, we need to recreate the client.
 		e.metricClient, err = e.newMetricClient(e.ctx, e.opts)
 		if err != nil {
 			return fmt.Errorf("create metric client: %w", err)
 		}
 	}
 
+	// Updates series cache settings; those require extra care on cache entries.
+	if !e.matchers.Equals(toApply) {
+		e.matchers = toApply
+		e.seriesCache.setMatchers(e.matchers)
+	}
 	if labelsChanged {
 		e.externalLabels = lset
 		// New external labels possibly invalidate the cached series conversions.
 		e.seriesCache.forceRefresh()
 	}
-
 	return nil
 }
 
@@ -557,12 +592,17 @@ func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample, exemp
 		return
 	}
 
-	metadata = e.wrapMetadata(metadata)
-
 	e.mtx.Lock()
+	if !e.configured {
+		// ApplyConfig must be called at least once, otherwise we risk a race between flag and runtime config change on start.
+		e.mtx.Unlock()
+		return
+	}
 	externalLabels := e.externalLabels
 	start, end, ok := e.lease.Range()
 	e.mtx.Unlock()
+
+	metadata = e.wrapMetadata(metadata)
 
 	if !ok {
 		exemplarsDropped.WithLabelValues("not-in-ha-range").Add(float64(len(exemplarMap)))
@@ -1028,6 +1068,21 @@ func (m *Matchers) Set(s string) error {
 
 func (m *Matchers) IsCumulative() bool {
 	return true
+}
+
+// Equals returns true if m is the same as other matchers. Order matters.
+//
+// NOTE: In practice order semi-matter, as the match result should be similar, but
+// things are faster or slower depending on the order. For efficiency of this function
+// and deterministic efficiency of matching, we assume order matter.
+func (m *Matchers) Equals(other Matchers) bool {
+	return slices.EqualFunc(*m, other, func(msel, osel labels.Selector) bool {
+		return slices.EqualFunc(msel, osel, func(matcher, otherMatcher *labels.Matcher) bool {
+			return matcher.Type == otherMatcher.Type &&
+					matcher.Name == otherMatcher.Name &&
+					matcher.Value == otherMatcher.Value
+		})
+	})
 }
 
 func (m *Matchers) Matches(lset labels.Labels) bool {
--- a/google/export/export_test.go
+++ b/google/export/export_test.go
@@ -16,15 +16,18 @@ package export
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
+	"slices"
+	"sort"
+	"strings"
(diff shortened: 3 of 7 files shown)


hsmatulis (approved):

google/export/export.go
```diff
@@ -442,6 +443,7 @@ const (
 // ApplyConfig updates the exporter state to the given configuration. For the
 // Prometheus fork config's google_cloud entry, ExporterOpts might be
 // changed and applied to the exporter, potentially recreating the metric client.
+// NOTE: Runtime configuration will not override disable/disableAuth options from flags.
```
bwplotka, line 458: This is quite annoying thing I found (already the case before my PR, just adding note).

Essentially collector will fail on start without good auth and one could disable auth to "ignore" this error. Then it's kind of expected you can configure things via fork Prometheus config with correct credentials and force auth -- this is currently not supported (perhaps credential setting should "enable auth"). This gap is really blocking self-deploy GMP operator use cases (it requires some modifications to manifests).

Anyway, nothing now and we can fix in the next PR.

bwplotka: This is the likely root cause for some friction in using explicit credentials in our tests https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1756#issuecomment-3292763078

dashpole (approved):

google/config/config.go
```diff
@@ -14,7 +14,19 @@ type GoogleCloudExportConfig struct {
...
+	// EnableMatch makes export to use Match configuration.
+	// This was added due to a period of time when Match was not working.
+	// Re-enabling (fixing) it by default would be breaking change, so opt-in
+	// flow was internally added. Googlers, see go/gmp:matchstuck for details.
+	EnableMatch bool `yaml:"match,omitempty"`
+	// Match offers runtime change-able control of the "matchOneOf" filtering.
+	// This filtering will skip certain series from entering the series cache (from export).
+	// This data will still be ingested into Prometheus local storage.
+	//
+	// This overrides all --export.match flags from CLI and EXTRA_ARGS.
```
pintohutch: IIUC all config file parameters that have a corresponding EXTRA_ARGS flag take precedence, correct?

google/export/setup/setup.go
```diff
@@ -135,9 +135,8 @@ func ExporterOptsFlags(a *kingpin.Application, opts *export.ExporterOpts) {
 		Default(opts.Cluster).
 		StringVar(&opts.Cluster)
 
-	a.Flag("export.match", `A Prometheus time series matcher. Can be repeated. Every time series must match at least one of the matchers to be exported. This flag can be used equivalently to the match[] parameter of the Prometheus federation endpoint to selectively export data. (Example: --export.match='{job="prometheus"}' --export.match='{__name__=~"job:.*"})`).
-		Default("").
-		SetValue(&opts.Matchers)
+	a.Flag("export.match", `DEPRECATED: A Prometheus time series matcher. Can be repeated. Every time series must match at least one of the matchers to be exported. This flag can be used equivalently to the match[] parameter of the Prometheus federation endpoint to selectively export data. External labels are excluded from matching. (Example: --export.match='{job="prometheus"}' --export.match='{__name__=~"job:.*"})`).
```
pintohutch: Why are we marking it as deprecated in this PR?

google/export/series_cache.go
```diff
@@ -169,6 +169,36 @@ func (c *seriesCache) run(ctx context.Context) {
...
+			continue
+		}
+		if dropped && !e.dropped {
+			// Series was matching before and is now dropped.
+			e.dropped = true
+			// Release resources.
+			// This is best-effort -- only needed for efficiency, not correctness.
+			c.pool.release(e.protos.gauge.proto)
+			c.pool.release(e.protos.cumulative.proto)
+			e.protos = cachedProtos{}
```
pintohutch, line 187: Why are the protos and metadata being reset here? Is it because we may have seen a labelset change?

google/export/series_cache.go
```diff
@@ -169,6 +169,36 @@ func (c *seriesCache) run(ctx context.Context) {
 	}
 }
 
+// setMatchers changes internal series cache matcher.
+func (c *seriesCache) setMatchers(matchers Matchers) {
```
pintohutch, line 165: We should update the guidance in https://github.com/GoogleCloudPlatform/prometheus/blob/eb9a01e913311c48493f9d828622d4933616c0d3/google/export/series_cache.go#L118-L122, given we're permitting runtime change now.

google/export/series_cache.go
```diff
@@ -169,6 +169,36 @@ func (c *seriesCache) run(ctx context.Context) {
...
+
+	c.matchers = matchers
+
+	// Go through all entries and update their dropped status.
+	for _, e := range c.entries {
+		dropped := !c.matchers.Matches(e.lset)
+		if !dropped && e.dropped {
+			// Series was dropped before and is now matching.
+			e.dropped = false
+			e.nextRefresh = 0 // Force refresh to re-populate it.
```
pintohutch, line 177: I don't know if https://github.com/GoogleCloudPlatform/prometheus/blob/eb9a01e913311c48493f9d828622d4933616c0d3/google/export/series_cache.go#L127 is worth exploring or if setting to 0 is fine.

google/export/export_test.go
```diff
@@ -482,54 +437,226 @@ func TestApplyConfig(t *testing.T) {
...
+			expectedSamplesSeries: []gcmSampleLabels{
+				{metric: []string{"ref", "2"}, resource: resourceWithProject("project-test")},
+				{metric: []string{"ref", "3"}, resource: resourceWithProject("project-test")},
+			},
 		},
-		GoogleCloud: gcmconfig.GoogleCloudConfig{
-			Export: gcmconfig.GoogleCloudExportConfig{
-				Compression: "gzip",
+		{
+			name: "Filtering change (noop again)",
```
pintohutch, line 583: Great tests. Was there a particular scenario you were testing by re-enabling the no-op and force-filtering test cases here at the end?

pintohutch (commented):
LGTM % questions/comments

We could also use a `/prombench` run to ensure we don't see any regressions in steady-state performance.

pintohutch:
I couldn't see any results there... maybe our prombench setup needs fixing...

bwplotka:
I'd be very surprised if it worked given we didn't touch it for ~year.

It might be only one thing though: 

image

However, @pintohutch should we spend time fixing it now vs progressing on matcher stuckness? Is there anything you are are worried about in terms of perf in this PR? I am not adding any state and we don't filter on prombench.

google/export/setup/setup.go
bwplotka: Yes, but you are right I wasn't clear about it on the PR title sorry (I mentioned in the design doc though). 

We did start work towards this direction, so I thought we early warnings slowly. It will be gone with unfork too.

Would you suggest we do NOT do this in this PR or as part of this fix? We already had deprecation mentioned in many places e.g. 

Image

google/config/config.go
bwplotka: Yes, moved this comment to the top Config section, great observation, thanks 👍🏽

google/export/export_test.go
bwplotka: Yes, but not critical cases. Like some kind of reverts e.g.
*  operator reverted to pre-fix config state but collector not restarted.
*  if we go with some forcing field and user accidentally removes that forcing field for some reasons (e.g. terraform) - we probably want to "preserve" old config.

Now when I look at it again it does feel weird that this particular case does not revert to empty filtering making `enableMatch` a clear companion to `match` option, so when enable is false it goes back to no filtering.... feels easier to reason about, but also..  easier to break yourself?

In the current state `enableMatch` is a  (A) switch for changes, so it's not declarative, but rather make exporter consider options. I think it might be safer for now. Updated the config `GoogleCloudExportConfig.EnableMatch` commentary to be clear on this.

```
// EnableMatch makes export to use Match configuration for a given configuration
// update.
```

...but feedback welcome if you prefer declarative approach -- operator could guard from some mistakes too.

If we do declarative (B), this test would look as follows:

```go
	{
			name: "Filtering change (noop matchers which means we should ignore config matchers)",
			toApply: &config.Config{
				GlobalConfig: config.GlobalConfig{
					ExternalLabels: labels.FromStrings(KeyProjectID, "project-test"),
				},
				GoogleCloud: gcmconfig.GoogleCloudConfig{
					Export: gcmconfig.GoogleCloudExportConfig{
						Compression: "gzip",
						EnableMatch: false, // noop.
						Match:       []string{`{ref="1"}`}, // Skip 2 and 3.
					},
				},
			},
			expectedSamplesSeries: []gcmSampleLabels{
				{metric: []string{"ref", "1"}, resource: resourceWithProject("project-test")},
				{metric: []string{"ref", "2"}, resource: resourceWithProject("project-test")},
				{metric: []string{"ref", "3"}, resource: resourceWithProject("project-test")},
			},
		},
```

bwplotka (commented):
Thanks for review!

As per above comment, I am not fixing prombench, but instead prioritize further parts of go/gmp:matchstuck

google/export/series_cache.go
bwplotka: oh wow, nice catch!

You made me worries that I forgot about some invariant but this works well because e.dropped is updated (and nextRefresh is set to 0). Killed this confusing comment.

google/export/series_cache.go
bwplotka: Using setNextRefresh would be disastrous (dropped are not populated) - I double check and our tests catch this 👍🏽

google/export/series_cache.go
bwplotka: No -- as the comment says it's to save precious bytes in memory.

google/config/config.go
```diff
@@ -11,10 +14,23 @@ type GoogleCloudConfig struct {
...
 
-	// Deprecated: This option no longer works, see https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1688
+	// EnableMatch makes export to use Match configuration for a given configuration
+	// update.
+	//
+	// This was added due to a period of time when Match was not working.
+	// Re-enabling (fixing) it by default would be breaking change, so opt-in
+	// flow was internally added. Googlers, see go/gmp:matchstuck for details.
+	EnableMatch bool `yaml:"match,omitempty"`
+	// Match, combined EnableMatch, controls the ExporterOpts.Matchers setting.
```
pintohutch: nit: "...combined with..."

google/export/setup/setup.go
pintohutch: Yea my vote is to add the deprecated notices separately, and focus on the enableMatch functionality in this PR.

google/export/export.go
pintohutch: More generally, what takes precedence. Flags? Or config?

i.e. Is the auth flags taking precedence WAI or a bug?

google/export/export.go
```diff
@@ -477,20 +478,38 @@ func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
 		}
 	}
 
-	// If changed, or we're calling this for the first time, we need to recreate the client.
+	matchersChanged := false
+	// Matchers are deprecated, but we still maintain them for backward compatibility, behind an opt-in flag.
+	if cfg.GoogleCloud.Export.EnableMatch {
+		var selectors []labels.Selector
```
pintohutch: We have a type for this in [`Matchers`](https://github.com/GoogleCloudPlatform/prometheus/blob/gmpmatchfix/google/export/export.go#L1030). Could we doing something like:
```go
		var selectors Matchers
		for _, match := range cfg.GoogleCloud.Export.Match {
			if err := selectors.Set(match); err != nil {
				return fmt.Errorf("setting selector with match %q: %w", match, err)
			}
		}
```

google/export/export.go
```diff
@@ -477,20 +478,38 @@ func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
...
+		for _, match := range cfg.GoogleCloud.Export.Match {
+			selector, err := parser.ParseMetricSelector(match)
+			if err != nil {
+				return fmt.Errorf("invalid metric matcher %q: %w", match, err)
+			}
+			selectors = append(selectors, selector)
+		}
+		matchersChanged = !reflect.DeepEqual(e.opts.Matchers, selectors)
```
pintohutch: nit (non-blocking): if we wanted to avoid reflection, could it be worth writing a quick `Matchers.Equals()` method?

We could also do that later as an optimization.

google/export/series_cache.go
```diff
@@ -133,10 +130,10 @@ func (e *seriesCacheEntry) setNextRefresh() {
 }
 
 func newSeriesCache(
-	logger log.Logger,
-	reg prometheus.Registerer,
-	metricTypePrefix string,
-	matchers Matchers,
+		logger log.Logger,
```
pintohutch: nit: why the extra indent?

google/export/series_cache.go
```diff
@@ -169,6 +166,36 @@ func (c *seriesCache) run(ctx context.Context) {
 	}
 }
 
+// setMatchers changes internal series cache matcher.
+func (c *seriesCache) setMatchers(matchers Matchers) {
```
pintohutch, line 165: suggestion: consider adding a unit test `TestSetMatchers` to series_cache.go to quickly assert the correct behavior.

google/export/export_test.go
```diff
@@ -482,54 +437,226 @@ func TestApplyConfig(t *testing.T) {
...
+			expectedCacheEntriesRefreshed: []storage.SeriesRef{1}, // Expect refresh for a series that's now matching.
+			expectedSamplesSeries: []gcmSampleLabels{
+				{metric: []string{"ref", "1"}, resource: resourceWithProject("project-test")},
+				{metric: []string{"ref", "2"}, resource: resourceWithProject("project-test")},
+				{metric: []string{"ref", "3"}, resource: resourceWithProject("project-test")},
+			},
+		},
+	} {
+		t.Run(tcase.name, func(t *testing.T) {
+			t.Helper()
```
pintohutch: question: is this actually a test helper function?

It looks like you're doing actual assertions in this scope. Same for L587 and L593.

google/export/export_test.go
```diff
@@ -482,54 +437,226 @@ func TestApplyConfig(t *testing.T) {
...
+
+			defer metricServer.clear()
+
+			if tcase.expectClientReload {
+				e.newMetricClient = func(_ context.Context, opts ExporterOpts) (metricServiceClient, error) {
+					t.Helper()
+					require.Equal(t, tcase.toApply.GoogleCloud.Export.Compression, opts.Compression)
```
pintohutch, line 684: nit (non-blocking): we have typically used [go-cmp](https://github.com/google/go-cmp) for equality assertions.

For example, see https://github.com/GoogleCloudPlatform/prometheus/blob/gmpmatchfix/google/export/export_test.go#L325-L327.

pintohutch, line 684: question: why is the compression settings validated here in the newMetricClient function?

Why not do it below?

google/export/export_test.go
pintohutch: Oh I see - yea should EnableMatch decide whether your previous matchers are used or not...

Really great callout. I would actually elaborate more in the docstring for that case and make it explicit.

OTOH I wonder if EnableMatch be an enum to be declaritive? i.e.
- No-op (leave filtering state alone)
- False - disable any notion of matching I've ever done or played around with.
- True - enforce the currently-specified matchers.

WDYT?

pintohutch (commented):
Great conversations - some more feedback for you (apologies for the delay)

google/export/export_test.go
bwplotka: Can change to that, although I prefer explicit enum for those vs magic `*bool` shenanigans

As a result, more flexible for future, documented and clear.

google/export/export_test.go
bwplotka: fair

google/export/export_test.go
bwplotka: Quick answer is to ensure correct compression is propagated to newMetricClient func correctly. Will double check if there's a better place for it.

google/export/export_test.go
pintohutch: SGTM

google/export/export_test.go
pintohutch: Oh right ok - yea that makes sense then.

google/export/export.go
bwplotka: In terms of what credentials to use -- config, so WAI.

But it does not matter for case I mention, essentially there's NO runtime config option for turning on explicitly disabled (via flag) auth/export. This is sort of missing feature which make things more reliable/functional. 

Something to add later, adding issue about it: https://github.com/GoogleCloudPlatform/prometheus/issues/261

pintohutch: Ack thanks

google/export/export_test.go
bwplotka: Sorry for not updating this -- I just found more reasons why to pick 2 states instead of 3. Let's discuss and decide on the document.

google/export/export.go
bwplotka: done

google/export/export_test.go
bwplotka: FIxed, used your *bool idea.

google/export/export_test.go
bwplotka: Yea, Prometheus uses `require` so we could simplify our code. WDYT?

bwplotka: Hm, It would be easier to backport if we don't use require I guess... let me fix.

bwplotka: eh.. 🙈 it will be painful... will do tomorrow if needed. Otherwise I addressed everything else

image

bwplotka:
PTAL, this should be now good to @pintohutch 

I did add one more tiny change as a safeguard, please check the description note: https://github.com/GoogleCloudPlatform/prometheus/pull/259#issue-3409411498

dashpole (approved):

google/export/export_test.go
```diff
@@ -482,54 +462,285 @@ func TestApplyConfig(t *testing.T) {
...
+					ExternalLabels: labels.FromStrings(KeyProjectID, "project-test"),
+				},
+				GoogleCloud: gcmconfig.GoogleCloudConfig{
+					Export: gcmconfig.GoogleCloudExportConfig{
+						Compression: "gzip",
+						// EnableMatch is now nil.
+					},
+				},
+			},
+			expectedCacheEntriesRefreshed: []storage.SeriesRef{1}, // Expect refresh for series that're now matching.
```
pintohutch, line 595: Why would `ref: 1` be matching now?

IIUC we went from skipping `ref: 1` to a `enableMatch: nil`, which shouldn't change from the previous state right?

google/export/export_test.go
```diff
@@ -482,54 +462,285 @@ func TestApplyConfig(t *testing.T) {
...
+		{
+			name: "Filtering change (force empty)",
+			toApply: &config.Config{
+				GlobalConfig: config.GlobalConfig{
+					ExternalLabels: labels.FromStrings(KeyProjectID, "project-test"),
+				},
+				GoogleCloud: gcmconfig.GoogleCloudConfig{
+					Export: gcmconfig.GoogleCloudExportConfig{
+						Compression: "gzip",
+						EnableMatch: proto.Bool(true),
```
pintohutch, line 666: Shouldn't this clear to matching nothing?

I would've thought no cache entries were refreshed (i.e. `nextRefresh` != 0) and samples in the metric server would be empty...but maybe I'm missing something.

google/export/export_test.go
bwplotka: Because as in our config description and design *bool == nil means noop/our current behaviour, so config ignored and initial default filters (flags if any) applied.

Initial filter skips 3, so this is what we see, 1 is not matching again. The test case description also says "noop", so back to "config is noop again". 

Why do you expect going from

 * enableMatch = true  && match=[skip 1]
 to
 * enableMatch = nil 

to keep skip 1 matching?

Config is fully declarative now and it makes it simple to reason about. If you expected imperative "diff" config logic, so config change is either applicable if enableMatch=true or old config should be used, then it makes things inconsistent --imagine user doing this and then collector restarts and forgets last state etc... this will get distributed across 1000 nodes and it's likely not what we want?

Glad you are asking those questions, sooo many nuances (: But I tried to explain it everywhere: 3 state *bool with declarative, static behaviour.

google/export/export_test.go
bwplotka: EnableMatch = true means current matchers are applied from config. Config matchers are empty meaning apply-all. This is mentioned in comment in config, flag help and it's designed to match Prometheus federation `match[]` logic (empty means matching all).  Likely not something we want to suddenly change.

It MIGHT be a good idea to do in another PR "disable export" runtime option if we want though if you want to have a non hacky way to disable export for GMP... but is it a good use case?

google/export/export_test.go
pintohutch: Ah yes - I was initially thinking:
nil == noop == don't make any changes to whatever is happening now _or_ going forward.

But I think your logic makes more sense.

I also missed that the initial flag-mocked value matches 1, so I get the flow now.

google/export/export_test.go
pintohutch: > EnableMatch = true means current matchers are applied from config. Config matchers are empty meaning apply-all. This is mentioned in comment in config, flag help and it's designed to match Prometheus federation match[] logic (empty means matching all). Likely not something we want to suddenly change.

Makes sense.

> It MIGHT be a good idea to do in another PR "disable export" runtime option if we want though if you want to have a non hacky way to disable export for GMP... but is it a good use case?

No I think this is ok. Though maybe a suggestion(nit): add a test case for `EnableMatch: true` that drops everything?

pintohutch (approved):
LGTM % nit

Thanks for doing this!

google/export/export_test.go
bwplotka: Done!

google/export/export_test.go
bwplotka: Ok, I got positive opinion from @dashpole on doing this, so merging without changing to more verbose form.

PR #267, opened 2025-11-24
feat: support info metrics

bernot-dev:
Add support for INFO metrics by converting them to Gauges.

### DIFF
--- a/google/export/series_cache.go
+++ b/google/export/series_cache.go
@@ -446,6 +446,14 @@ func (c *seriesCache) populate(ref storage.SeriesRef, entry *seriesCacheEntry, e
 			metric_pb.MetricDescriptor_DOUBLE,
 		)
 
+	case model.MetricTypeInfo:
+		protos.gauge = newSeries(
+			c.getMetricType(metricName, gcmMetricSuffixGauge, gcmMetricSuffixNone),
+			metadata.Help,
+			metric_pb.MetricDescriptor_GAUGE,
+			metric_pb.MetricDescriptor_DOUBLE,
+		)
+
 	case model.MetricTypeUnknown:
 		protos.gauge = newSeries(
 			c.getMetricType(metricName, gcmMetricSuffixUnknown, gcmMetricSuffixNone),
--- a/google/export/series_cache_test.go
+++ b/google/export/series_cache_test.go
@@ -19,12 +19,80 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb/record"
+	metric_pb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredres_pb "google.golang.org/genproto/googleapis/api/monitoredres"
 	"google.golang.org/protobuf/testing/protocmp"
 )
+
+func TestSeriesCache_populate_Info(t *testing.T) {
+	cache := newSeriesCache(nil, nil, MetricTypePrefix)
+
+	// Mock getLabelsByRef to return labels for our info metric
+	ref := storage.SeriesRef(1)
+	metricName := "test_info"
+	lset := labels.FromStrings("__name__", metricName, "job", "test_job", "instance", "test_instance", "version", "1.2.3")
+	cache.getLabelsByRef = func(r storage.SeriesRef) labels.Labels {
+		if r == ref {
+			return lset
+		}
+		return labels.EmptyLabels()
+	}
+
+	// Prepare entry
+	entry := &seriesCacheEntry{}
+
+	// Metadata function returning Info type
+	mdFunc := func(m string) (MetricMetadata, bool) {
+		if m == metricName {
+			return MetricMetadata{
+				Metric: metricName,
+				Type:   model.MetricTypeInfo,
+				Help:   "info help",
+			}, true
+		}
+		return MetricMetadata{}, false
+	}
+
+	// Call populate
+	// Provide required external labels (project_id, location) to pass extractResource
+	externalLabels := labels.FromStrings("project_id", "my-project", "location", "us-central1")
+	err := cache.populate(ref, entry, externalLabels, mdFunc)
+	if err != nil {
+		t.Fatalf("populate failed: %v", err)
+	}
+
+	// Verify
+	if entry.protos.gauge.proto == nil {
+		t.Fatal("expected gauge proto to be populated for Info metric")
+	}
+	if entry.protos.cumulative.proto != nil {
+		t.Error("expected cumulative proto to be nil for Info metric")
+	}
+
+	p := entry.protos.gauge.proto
+
+	// Check MetricKind
+	if p.MetricKind != metric_pb.MetricDescriptor_GAUGE {
+		t.Errorf("expected MetricKind GAUGE, got %v", p.MetricKind)
+	}
+	// Check ValueType
+	if p.ValueType != metric_pb.MetricDescriptor_DOUBLE {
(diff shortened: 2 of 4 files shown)


google/export/series_cache.go
```diff
@@ -446,6 +446,14 @@ func (c *seriesCache) populate(ref storage.SeriesRef, entry *seriesCacheEntry, e
 			metric_pb.MetricDescriptor_DOUBLE,
 		)
 
+	case model.MetricTypeInfo:
+		protos.gauge = newSeries(
+			c.getMetricType(metricName, gcmMetricSuffixGauge, gcmMetricSuffixNone),
```
bwplotka, line 451: One question to answer before we move forward: Should we reuse gauge suffix`gcmMetricSuffixGauge` or add some `/info` suffix and if that decision even matter now (e.g. we can easily change later). Gauge would totally work for now. The only reason to do `info` is to be forward compatible with potential special handling of info in PromQL later on.  Similar for stateset.

FYI: the definitions of those rare types are here: https://prometheus.io/docs/specs/om/open_metrics_spec/#info

I think it's fine to have gauge for both now, because:
* There are no clear plans how to have special PromQL handling for info and stateset. There are small discussions on some storage efficiency gains for native info/stateset samples, but that's not impacting us. 
* If, ever, we will make breaking changes to Prometheus to suddenly treat info and stateset specially on PromQL it will be a big thing and in 4.0 Prometheus, soo.. YAGNI.
* Otel collector is already assuming a "sum" (gauge) AFAIK 

 @lyanco @realschwa @dashpole  to double check my thinking (:

FYI: There's some discussions to add info and stateset natively to SDK (https://cloud-native.slack.com/archives/CC6CPDEJV/p1762847796024439), so they will be more popular, but again, will be treated as a gauge on consumption.

bwplotka (commented):
That's generally should be it, but also test will tell us more.

I'd recommend adding one to https://github.com/GoogleCloudPlatform/prometheus/blob/release-2.53.5-gmp/google/internal/promqle2etest/gcm_test.go

Since SDK does not support info metrics you might need some slight change in how we setup test framework to inject custom text format 🤔 that might be the main work here. If I were you I would attempt to extend (you are welcome to contribute too!) https://github.com/prometheus/compliance/blob/7667b56d892fbb4f4252287e50bbad6d2a8f6893/promqle2e/promqle2e.go#L435 so it accepts []string with OpenMetrics raw TEXT to show (and change) on every scrape and expect some result.

This would benefit the `promqle2e` and our environment, but if you find a different way to test this (e.g. extending our side local export GCM helper: https://github.com/GoogleCloudPlatform/prometheus/blob/3a83a91a9e4fd25a9fa365a89765d5c08e65c523/google/internal/promqle2etest/backend_export_gcm.go#L158)

ghost (changes requested):
Jugnkjgy

google/export/series_cache.go
lyanco: +1 to YAGNI - gauge seems fine for now and it matches other info-like metrics that already exist (e.g. the ones from ksm).

Do these metrics typically end in _info? If so we can always special case metrics that end in _info/gauge if we ever need to. Or later on if /info becomes necessary we could always change how we write these metrics to include a /info suffix, or double-write (cardinality is not high for these, so cost isn't either), or something.

bernot-dev: The [info suffix is specified for OpenMetrics](https://prometheus.io/docs/specs/om/open_metrics_spec/#info-1). Not sure how well that translates to real-world usage.

bwplotka: > Do these metrics typically end in _info? 

Yes, but it's just a convention, although I think we keep is OpenMetrics MUST requirement.

Nevertheless sounds like a conclusion is that it is safe.

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
 	"github.com/prometheus/prometheus/tsdb/record"
+	metric_pb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredres_pb "google.golang.org/genproto/googleapis/api/monitoredres"
 	"google.golang.org/protobuf/testing/protocmp"
 )
 
+func TestSeriesCache_populate_Info(t *testing.T) {
+	cache := newSeriesCache(nil, nil, MetricTypePrefix)
+
+	// Mock getLabelsByRef to return labels for our info metric
```
bwplotka:
```suggestion
	// Mock getLabelsByRef to return labels for our info metric.
```

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
+	metricName := "test_info"
+	lset := labels.FromStrings("__name__", metricName, "job", "test_job", "instance", "test_instance", "version", "1.2.3")
+	cache.getLabelsByRef = func(r storage.SeriesRef) labels.Labels {
+		if r == ref {
+			return lset
+		}
+		return labels.EmptyLabels()
+	}
+
+	// Prepare entry
```
bwplotka:
```suggestion
	// Prepare entry.
```

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
+	entry := &seriesCacheEntry{}
+
+	// Metadata function returning Info type
```
bwplotka:
```suggestion
	// Metadata function returning Info type.
```

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
+				Metric: metricName,
+				Type:   model.MetricTypeInfo,
+				Help:   "info help",
+			}, true
+		}
+		return MetricMetadata{}, false
+	}
+
+	// Call populate
+	// Provide required external labels (project_id, location) to pass extractResource
```
bwplotka:
```suggestion
	// Populate cache. Provide required external labels (project_id, location).
```

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
+	externalLabels := labels.FromStrings("project_id", "my-project", "location", "us-central1")
+	err := cache.populate(ref, entry, externalLabels, mdFunc)
+	if err != nil {
+		t.Fatalf("populate failed: %v", err)
+	}
+
+	// Verify
```
bwplotka:
```suggestion
	// Verify the cache entry.
```

google/export/series_cache_test.go
```diff
@@ -19,13 +19,81 @@ import (
...
+	if entry.protos.gauge.proto == nil {
+		t.Fatal("expected gauge proto to be populated for Info metric")
+	}
+	if entry.protos.cumulative.proto != nil {
+		t.Error("expected cumulative proto to be nil for Info metric")
+	}
+
+	p := entry.protos.gauge.proto
+
+	// Check MetricKind
```
bwplotka: ditto (do we need those comments?)

google/export/transform_test.go
```diff
@@ -143,6 +143,47 @@ func TestSampleBuilder(t *testing.T) {
...
+							"project_id": "example-project",
+							"location":   "europe",
+							"cluster":    "foo-cluster",
+							"namespace":  "",
+							"job":        "job1",
+							"instance":   "instance1",
+						},
+					},
+					Metric: &metric_pb.Metric{
+						Type:   "prometheus.googleapis.com/metric1_info/gauge",
```
bwplotka, line 171: Looks the same as our lovely `prometheus.googleapis.com/gotk_resource_info/gauge` thing we try to fix this for 👍🏽

bwplotka (approved):
Thanks.

LGTM, modulo tiny commentary nits (keep it a full English sentence, please).

I wished we could test it e2e with GCM, but this is good enough, we can do e2e test later and manually deploy latest GMP with KSM with info metric.

Added: https://github.com/GoogleCloudPlatform/prometheus/issues/281

bwplotka:
NOTE: Same work is needed for `stateset`, but low prio.

PR #271, opened 2025-12-15
fix release-2.45.3-gmp vulnerabilities; remove vendor

bwplotka:
This fixes vulnerabilities. 

Unfortunately, on the way our old way of building UI also started to fail (new image upgraded node?): https://github.com/GoogleCloudPlatform/prometheus/actions/runs/20249220559/job/58137089941

As a result, it will be simpler to port 2.53.5 Prometheus fork artifacts.


Dockerfile.google
```diff
@@ -1,29 +1,72 @@
-ARG IMAGE_BUILD_NODEJS=launcher.gcr.io/google/nodejs
```
pintohutch, line 1: Could we get away with just:
`ARG IMAGE_BUILD_NODEJS=node:17.9.0-buster`?

and removing the `install_node` command on line 12?

And leave everything else alone except for the go bump to address the vulnz?

bwplotka: I can try, but I thought it will be easier and more consistent to 2.53.5 way of thins.

bwplotka: Trying on another PR: https://github.com/GoogleCloudPlatform/prometheus/pull/272

error so far  https://github.com/GoogleCloudPlatform/prometheus/actions/runs/20275550076/job/58223081923?pr=272

image

bwplotka:
Simpler fix: https://github.com/GoogleCloudPlatform/prometheus/pull/272

PR #282, opened 2026-02-12
feat: add stateset support

### LINKED ISSUE #490: export: Add support for Info, Stateset and gaugehistogram, metrics (new OM metric types).
Prometheus supports [OM metric types](https://github.com/prometheus/prometheus/blob/1ea477f4bcd04c580e51711dca346a06421b311c/model/textparse/interface.go#L115) (passed in metadata) since [2018](https://github.com/prometheus/prometheus/commit/c271ef456a0b696a389f7e607472c8c3573eafd1). From a quick glance [export returns error when OM is used with new types](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/series_cache.go#L462), which is emits [debug log message](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/series_cache.go#L242). 

### AC:

* Export supports Info, Stateset and gaugehistogram (?) metric types 

Alternatively we could instrument that error with the metric to know the scale of the problem.

bwplotka:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/490

Let's fix statesets which tend to be also used in OM 1 https://prometheus.io/docs/specs/om/open_metrics_spec/#stateset-1 and are present in KSM (see https://github.com/GoogleCloudPlatform/prometheus-engine/issues/490#issuecomment-1940851286)

Shamelessly copying @bernot-dev testability patterns from https://github.com/GoogleCloudPlatform/prometheus/pull/267

The only unsupported thing will be now `gaugehistogram` which I never seen in practice -- I don't think those are well supported in Prometheus OSS as well, nor if Monarch even supports those 🤔

### DIFF
--- a/google/export/export_test.go
+++ b/google/export/export_test.go
@@ -279,6 +279,22 @@ func TestExporter_wrapMetadata(t *testing.T) {
 			},
 			wantOK: true,
 		}, {
+			desc: "stateset metadata is returned as is",
+			mf: func(string) (MetricMetadata, bool) {
+				return MetricMetadata{
+					Metric: "some_stateset_metric",
+					Type:   model.MetricTypeStateset,
+					Help:   "stateset help",
+				}, true
+			},
+			metric: "some_stateset_metric",
+			want: MetricMetadata{
+				Metric: "some_stateset_metric",
+				Type:   model.MetricTypeStateset,
+				Help:   "stateset help",
+			},
+			wantOK: true,
+		}, {
 			desc: "not found metadata defaults to untyped",
 			mf: func(string) (MetricMetadata, bool) {
 				return MetricMetadata{}, false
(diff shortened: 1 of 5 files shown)


google/export/export_test.go
```diff
@@ -278,6 +278,22 @@ func TestExporter_wrapMetadata(t *testing.T) {
...
+				return MetricMetadata{
+					Metric: "some_stateset_metric",
+					Type:   model.MetricTypeStateset,
+					Help:   "stateset help",
+				}, true
+			},
+			metric: "some_stateset_metric",
+			want: MetricMetadata{
+				Metric: "some_stateset_metric",
+				Type:   model.MetricTypeInfo,
```
bernot-dev, line 293: This shouldn't be MetricTypeInfo, right?

bwplotka: Yes, CI just noticed this, fixed!

bernot-dev (approved):
LGTM

PR #295, opened 2026-03-19
chore: Update OpenTelemetry semantic conventions to v1.39.0

ridwanmsharif:
Fixes some of the broken CI tests


ridwanmsharif:
Should be a safe change, only attrs we care about are: https://github.com/GoogleCloudPlatform/prometheus/blob/e527b176767272cef99d4fb2a488c7754f80343c/tracing/tracing.go#L144-L150 and they haven't changed in the new versions

bernot-dev (approved):
LGTM

PR #322, opened 2026-07-01
test(export): add reproduction test cases for histogram distribution theft

bwplotka:
Add cases to TestSampleBuilder for:
- ungrouped (interleaved) histogram samples
- ungrouped (interleaved) histogram samples with first group incomplete
- ungrouped (interleaved) histogram samples with first group skipped due to new bucket
- grouped histogram samples
- grouped histogram samples with first group incomplete
- grouped histogram samples with first group skipped due to new bucket

NOTE: We expect tests to fail to show the issue.

TAG=agy
CONV=8f508481-de1c-4e6b-ad3a-718d089a2fbe

PR #323, opened 2026-07-01
fix(export): Fix histogram theft bug on interleaved and incomplete histograms

bwplotka:
> [!IMPORTANT]
> This also fixes incomplete histogram edge cases for non-interleaved series, however the fix could be simpler if we don't support interleaved series.

## Summary

When histogram samples in a scrape batch are ungrouped or interleaved across different series label sets for the same metric family (such as from non-compliant sources like [Kong/kong#14925](https://github.com/Kong/kong/issues/14925)), the existing `buildDistribution` logic in `sampleBuilder` returned prematurely as soon as any single histogram in the cache completed (`dist.complete()`).

This premature return caused two critical bugs:
1. **Histogram Distribution Theft**: Because the completed distribution was returned immediately without strictly binding emission to each specific series' metadata, completed distributions could be attached to the wrong series label sets.
2. **Sample Loss & State Confusion**: Returning early left the remaining interleaved samples in the contiguous block unprocessed or lost. Furthermore, incomplete or skipped histogram series (e.g., when new bucket bounds are introduced across scrapes) could disrupt valid distributions.

### 1. Contiguous Block Processing (`google/export/transform.go`)
- **Replaced Single-Distribution Builder**: Replaced `buildDistribution` with `buildDistributions`, called from `next`. Instead of aborting on the first completed distribution, `buildDistributions` now consumes the entire contiguous block of samples for a histogram metric family name before emitting results.
- **Per-Series Metadata Caching**: Extended the cached `distribution` struct to store series metadata (`hash`, `proto`, and `lset`). When the block finishes, all completed distributions are built and emitted using their own cached metadata, guaranteeing deterministic ordering and accurate label attribution.
- **Incomplete & Skipped Histogram Handling**: Incomplete histograms (such as those missing a `+Inf` bucket) or skipped series are cleanly discarded at the end of the block without being emitted under incorrect label sets or causing sample loss for other series.
- **Memory & Allocation Optimizations**: Added `results []hashedSeries` to `sampleBuilder` to pool and reuse result slices across iterations without extra heap allocations. Added a `defer` cleanup block in `buildDistributions` to return cached distribution objects to the pool (`putDistribution`) and clear the map after processing each metric family.

### 2. Regression Testing (`google/export/transform_test.go`)
- Added comprehensive regression test cases in `TestSampleBuilder` (referencing internal regression tracking `b/516519320`):
  - Ungrouped (interleaved) histogram samples across multiple series.
  - Ungrouped (interleaved) histogram samples where the first group is incomplete.
  - Ungrouped (interleaved) histogram samples where the first group is skipped due to schema/bucket changes across scrapes (e.g., adding a new bucket in a subsequent scrape).
- Added deterministic slice sorting (`cmpopts.SortSlices`) to test assertions when comparing multiple emitted histogram series.

## Related Issues
- Regression tracking: `b/516519320`
- Upstream context: [Kong/kong#14925](https://github.com/Kong/kong/issues/14925)

### DIFF
--- a/google/export/transform.go
+++ b/google/export/transform.go
@@ -70,16 +70,16 @@ func discardExemplarIncIfExists(series storage.SeriesRef, exemplars map[storage.
 }
 
 type sampleBuilder struct {
-	series  *seriesCache
-	dists   map[uint64]*distribution
-	results []hashedSeries
+	series         *seriesCache
+	dists          map[uint64]*distribution
+	histResultsBuf []hashedSeries
 }
 
 func newSampleBuilder(c *seriesCache) *sampleBuilder {
 	return &sampleBuilder{
-		series:  c,
-		dists:   make(map[uint64]*distribution, 128),
-		results: make([]hashedSeries, 0, 4),
+		series:         c,
+		dists:          make(map[uint64]*distribution, 128),
+		histResultsBuf: make([]hashedSeries, 0, 4),
 	}
 }
 
@@ -135,6 +135,8 @@ func (b *sampleBuilder) next(metadata MetadataFunc, externalLabels labels.Labels
 	}
 	if c := entry.protos.cumulative; c.proto != nil {
 		if entry.metadata.Type == model.MetricTypeHistogram {
+			// The moment we see a histogram we build all histogram series for a given metric family.
+			// This is to ensure we guard ourselves from sources that interleave histogram series e.g. https://github.com/Kong/kong/issues/14925.
 			var histSeries []hashedSeries
 			var err error
 			histSeries, tailSamples, err = b.buildDistributions(
@@ -487,7 +489,7 @@ Loop:
 		return nil, samples[1:], errors.New("no sample consumed for histogram")
 	}
 
-	b.results = b.results[:0]
+	b.histResultsBuf = b.histResultsBuf[:0]
 	for _, dist := range b.dists {
 		if dist.complete() {
 			dp, err := dist.build(dist.lset)
@@ -506,11 +508,11 @@ Loop:
 						Value: &monitoring_pb.TypedValue_DistributionValue{DistributionValue: dp},
 					},
 				}}
-				b.results = append(b.results, hashedSeries{hash: dist.hash, proto: &ts})
+				b.histResultsBuf = append(b.histResultsBuf, hashedSeries{hash: dist.hash, proto: &ts})
 			}
 		}
 	}
-	return b.results, samples[consumed:], nil
+	return b.histResultsBuf, samples[consumed:], nil
 }
 
 func buildExemplars(exemplars []record.RefExemplar) []*distribution_pb.Distribution_Exemplar {
--- a/google/export/transform.go
+++ b/google/export/transform.go
@@ -158,16 +158,14 @@ func (b *sampleBuilder) next(metadata MetadataFunc, externalLabels labels.Labels
 			)
 			var v float64
 			resetTimestamp, v, ok = b.series.getResetAdjusted(storage.SeriesRef(sample.Ref), sample.T, sample.V)
+			// We may not have produced a value if:
+			//
+			//   1. It was the first sample of a cumulative and we only initialized  the reset timestamp with it.
 			if ok {
 				value = &monitoring_pb.TypedValue{
 					Value: &monitoring_pb.TypedValue_DoubleValue{DoubleValue: v},
 				}
 				discardExemplarIncIfExists(storage.SeriesRef(sample.Ref), exemplars, "counters-unsupported")
-			}
-			// We may not have produced a value if:
-			//
-			//   1. It was the first sample of a cumulative and we only initialized  the reset timestamp with it.
-			if value != nil {
 				//nolint:govet
 				ts := *c.proto
 
@@ -397,6 +395,9 @@ func (b *sampleBuilder) buildDistributions(
 	// series. We consume all contiguous samples for the metric and return all completed histogram series.
 	consumed := 0
 	defer func() {
+		// Since we build all distributions for a given metric family in a single
+		// buildDistributions call, we know we don't need to reuse cache elements anymore.
+		// Release them so we limit memory use.
 		for _, dist := range b.dists {
 			putDistribution(dist)
 		}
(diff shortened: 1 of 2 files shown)


google/export/transform.go
```diff
@@ -230,6 +224,10 @@ type distribution struct {
 	hasSum, hasCount, hasInfBucket bool
 	// Whether to not emit a sample.
 	skip bool
+
+	hash  uint64
```
bwplotka, line 226: This is allocating bigger structs, but if we trust AI benchmark it does not yield big difference. 

Assessment: https://github.com/GoogleCloudPlatform/prometheus/pull/324

bwplotka:
Given some risks, I wonder if we shouldn't release an unofficial image first for cx to try.

bwplotka:
Experimental image with this fix, captured in [release-2.53.5-gmp-histcompat](https://github.com/GoogleCloudPlatform/prometheus/tree/release-2.53.5-gmp-histcompat) branch:
 [us-east1-docker.pkg.dev/gpe-test-1/public/prometheus-engine/prometheus:v2.53.5-gmp-histcompat.0-gke.0](http://us-east1-docker.pkg.dev/gpe-test-1/public/prometheus-engine/prometheus:v2.53.5-gmp-histcompat.0-gke.0)

bwplotka:
Our version won't work, looks like upstream image broke compatibility https://screenshot.googleplex.com/6zKrakMeUURwb4t

bwplotka:
I have updated tests to show it also fixes "grouped" cases: https://github.com/GoogleCloudPlatform/prometheus/pull/323/commits/bab9196b359460200c6e02aba25740ab861d17d9

Repro test showing it fails: https://github.com/GoogleCloudPlatform/prometheus/pull/322

bernot-dev:
This seems like a good fix to me, but I would like to have a better idea about the actual performance impact, beyond AI speculation. It seems like there's some real risk that this could cause memory usage to increase noticeably for large customers.

google/export/transform.go
```diff
@@ -379,24 +380,27 @@ func isHistogramSeries(metric, name string) bool {
...
 	externalLabels labels.Labels,
 	metadata MetadataFunc,
-) (*distribution_pb.Distribution, int64, []record.RefSample, error) {
-	// The Prometheus/OpenMetrics exposition format does not require all histogram series for a single distribution
-	// to be grouped together. But it does require that all series for a histogram metric in general are grouped
-	// together and that buckets for a single histogram are specified in order.
-	// Thus, we build a cache and conclude a histogram complete once we've seen it's _sum series and its +Inf bucket
-	// series. We return for the first histogram where this condition is fulfilled.
+) ([]hashedSeries, []record.RefSample, error) {
+	// The Prometheus/OpenMetrics exposition format require all histogram series for a single distribution
```
dashpole: nit: Not sure if the Prometheus text format requires this at all: https://github.com/prometheus/docs/blob/main/docs/instrumenting/exposition_formats.md#grouping-and-sorting.  It just requires histogram buckets to be sorted, and for all series for a given metric name to be grouped together.

dashpole:
This looks correct to me.  I wondered if `convert_classic_histograms_to_nhcb` would be able to fix this for us, but it also has the same issue:

https://github.com/prometheus/prometheus/blob/9a6814cdfbb04825022877e4d4c0d518b3b1d9b3/model/textparse/nhcbparse.go#L202-L234

google/export/transform.go
bwplotka: Yes, as per https://github.com/Kong/kong/issues/14925

it's more of an OpenMetrics 1, 2 aspect, but we all rely on this in the ecosystem, as you noticed (NHCB)

bwplotka: I should remove "Prometheus" from here true

dashpole (approved):

google/export/transform.go
bwplotka: As expected benchmark was noop (samples not really created, flow was not showing full scrape loop)

google/export/transform_test.go
```diff
@@ -1622,6 +1622,628 @@ func TestSampleBuilder(t *testing.T) {
...
+									BucketCounts: []int64{0, 4},
+								},
+							},
+						},
+					}},
+				},
+			},
+		},
+		{
+			doc: "grouped histograms samples with the first group being incomplete",
```
bwplotka, line 2072: non-interleaved case that we fix with this change.

bwplotka:
[Benchmarks looks ok](https://github.com/GoogleCloudPlatform/prometheus/pull/327#issuecomment-4913134119). Kong source improved (interleaving, stable bucketing), but the same problem can occur for other customers and custom metric sources.

Notably this also fixes case of complete histogram for the same metric after an incomplete one: https://github.com/GoogleCloudPlatform/prometheus/pull/323/changes#r3542690086

WDYT @bernot-dev @pintohutch ?

google/export/transform.go
```diff
@@ -169,23 +164,22 @@ func (b *sampleBuilder) next(metadata MetadataFunc, externalLabels labels.Labels
...
-					StartTime: getTimestamp(resetTimestamp),
-					EndTime:   getTimestamp(sample.T),
-				},
-				Value: value,
-			}}
-			result = append(result, hashedSeries{hash: c.hash, proto: &ts})
+			// We may not have produced a value if:
+			//
+			//   1. It was the first sample of a cumulative and we only initialized  the reset timestamp with it.
+			if value != nil {
```
pintohutch: nit(non-blocking): value only looks set in the ok check above

Is it necessary for a separate check?

I'm wondering if the variable declarations of `value`, `resetTimestamp`, and `v` could be done in-line with everything falling through the `if ok` clause.

google/export/transform.go
```diff
@@ -379,24 +380,28 @@ func isHistogramSeries(metric, name string) bool {
...
+) ([]hashedSeries, []record.RefSample, error) {
+	// The OpenMetrics exposition format require all histogram series for a single distribution
+	// to be grouped together. However, Prometheus Text format do not restrict this
+	// (https://github.com/prometheus/docs/blob/main/docs/instrumenting/exposition_formats.md#grouping-and-sorting)
+	// and some sources interleaves series (e.g. https://github.com/Kong/kong/issues/14925).
+
+	// We build a cache and conclude a histogram complete once we've seen its _sum series and its +Inf bucket
+	// series. We consume all contiguous samples for the metric and return all completed histogram series.
 	consumed := 0
+	defer func() {
```
pintohutch, line 397: nit: comment on what the defer is doing and why

google/export/transform.go
```diff
@@ -475,23 +483,37 @@ Loop:
...
-		return dp, dist.resetTimestamp, samples[consumed:], nil
 	}
 	if consumed == 0 {
 		prometheusSamplesDiscarded.WithLabelValues("zero-histogram-samples-processed").Inc()
 		discardExemplarIncIfExists(storage.SeriesRef(samples[0].Ref), exemplars, "zero-histogram-samples-processed")
-		return nil, 0, samples[1:], errors.New("no sample consumed for histogram")
+		return nil, samples[1:], errors.New("no sample consumed for histogram")
+	}
+
+	b.histResultsBuf = b.histResultsBuf[:0]
```
pintohutch, line 494: nit: add comment on what this final loop is doing

google/export/transform.go
```diff
@@ -70,14 +70,16 @@ func discardExemplarIncIfExists(series storage.SeriesRef, exemplars map[storage.
...
+	histResultsBuf []hashedSeries
 }
 
 func newSampleBuilder(c *seriesCache) *sampleBuilder {
 	return &sampleBuilder{
-		series: c,
-		dists:  make(map[uint64]*distribution, 128),
+		series:         c,
+		dists:          make(map[uint64]*distribution, 128),
+		histResultsBuf: make([]hashedSeries, 0, 4),
```
pintohutch, line 82: anything magic about the number 4?

IIUC it's a rough balance between "dont allocate too much" and "allocate enough where we can handle a handful of interleaved series in a family", but please correct me if im wrong

pintohutch (approved):
Wow - really great work here - what a gnarly bug!

Added a few nits and questions, but LGTM otherwise.

IIUC for interleaved series in a metric family, the "histogram distribution theft" wasn't limited to buckets, but also time intervals, correct? Thus leading to the OOOs?

google/export/transform.go
bwplotka: That's correct, it's an arbitrary guess, but logical - feels like a good avg case

bwplotka:
> IIUC for interleaved series in a metric family, the "histogram distribution theft" wasn't limited to buckets, but also time intervals, correct? Thus leading to the OOOs?

Start time intervals strictly. Because we sent B series data as A series accidentally it can happen that B was started/reset way after A. Then when A is complete again we have OOO due to start time overlap. 

I can imagine more cases too e.g. We suddenly sent B series data as A series accidentally and B was started before A -- it can also trigger the same effect

bwplotka:
Addressed!

PR #328, opened 2026-07-09
Increase GCM test timeout to 20 minutes

bwplotka:
Increased timeout for GCM tests from 15m to 20m.


bwplotka:
I see the data on range query, so maybe it's just slower? https://screenshot.googleplex.com/AoMA5HZ7dgzWPGY

bernot-dev (approved):

bwplotka:
The problem was somewhere else, 15m is enough typically.

PR #335, opened 2026-07-21
build: optimize Dockerfile.google multi-platform compilation speed

bernot-dev:
Pin compiler stages (gobase and nodebase) to the native build host platform ($BUILDPLATFORM) to run CPU-bound tools natively instead of under QEMU emulation.

* Configure Go build to cross-compile to the target architecture ($TARGETARCH) using env variables.
* Export CC compiler variable in the cross-compilation block to ensure the correct CGO toolchain is visible to go build.
* Use nvm under a native base stage to support architectures where the upstream NodeJS image has no native variant.
* Update legacy ENV instructions to standard key=value syntax.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,3 +11,13 @@ data/
 
 /vendor
 
+# Ignore local compiled binaries
+/prometheus
+/promtool
+/promtool-benchmark
+
+# Ignore git repository details unless required
+.git/
+
+# Ignore local Node dependencies
+**/node_modules/
--- a/Dockerfile.google
+++ b/Dockerfile.google
@@ -4,29 +4,48 @@
 
 # For the lack of the other official Google nodejs image, we use serverless project
 # images to build the Prometheus frontent (https://cloud.google.com/docs/buildpacks/base-images).
-ARG IMAGE_BUILD_NODEJS=us-central1-docker.pkg.dev/serverless-runtimes/google-22/runtimes/nodejs22:latest@sha256:9e88442205b4c956ca4996c2be626db6ef412043182cdd620e741d0d5e14b6a6
+ARG IMAGE_BUILD_NODEJS=gke.gcr.io/debian-base:bookworm-v1.0.8-gke.9@sha256:fc22d178e95261ca59987ca396107c4bd8b6f90d4a64c7f387f22eeb4529546a
 ARG IMAGE_BUILD_GO=google-go.pkg.dev/golang:1.26.4@sha256:3444149d0a7e3f7cfb9c2db65f0f75676fe6ad04de3ce72674efb120c08dd1c1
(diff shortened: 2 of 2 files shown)


Dockerfile.google
```diff
@@ -86,7 +97,7 @@ ENV GOOS=${TARGETOS}
 RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
       apt install -y --no-install-recommends \
         gcc-aarch64-linux-gnu libc6-dev-arm64-cross; \
-      CC=aarch64-linux-gnu-gcc; \
+      export CC=aarch64-linux-gnu-gcc; \
     fi && \
```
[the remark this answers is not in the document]
bernot-dev: Cleanup is not necessary because it's an intermediate image.

Dockerfile.google
```diff
@@ -4,28 +4,39 @@
...
+
+# Create a script file sourced by both interactive and non-interactive bash shells
+ENV BASH_ENV="/root/.bash_env"
+RUN touch "${BASH_ENV}"
+RUN echo '. "${BASH_ENV}"' >> ~/.bashrc
+
+# Download and install nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.6/install.sh | PROFILE="${BASH_ENV}" bash
+COPY web/ui/.nvmrc .nvmrc
+RUN nvm install
```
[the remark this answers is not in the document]
bernot-dev: Going to stick with the recommended nvm Docker install steps.

.dockerignore
```diff
@@ -11,3 +11,13 @@ data/
 
 /vendor
 
+# Ignore local compiled binaries
```
bwplotka, line 14: Why? Those shouldn't exist 🤔

Dockerfile.google
```diff
@@ -84,9 +103,15 @@ ENV GOWORK=off
...
-        gcc-aarch64-linux-gnu libc6-dev-arm64-cross; \
-      CC=aarch64-linux-gnu-gcc; \
+      apt-get update && apt-get install -y --no-install-recommends \
+        gcc-aarch64-linux-gnu libc6-dev-arm64-cross && \
+      rm -rf /var/lib/apt/lists/*; \
+      export CC=aarch64-linux-gnu-gcc; \
+    elif [ "${TARGETARCH}" = "amd64" ] && [ "${BUILDARCH}" != "amd64" ]; then \
+      apt-get update && apt-get install -y --no-install-recommends \
+        gcc-x86-64-linux-gnu libc6-dev-amd64-cross && \
+      rm -rf /var/lib/apt/lists/*; \
```
bwplotka, line 113: Why we delete list of packages?

bwplotka (approved):
Let's try!

Dockerfile.google
bernot-dev: GCA was really insistent about it. It shouldn't hurt anything, but it probably doesn't help as much as GCA thinks. It slightly reduces the intermediate/cached image file sizes.

PR #349, opened 2026-08-04
tracing: remove explicit schema URL when building tracer provider resource (release-2.53.5-gmp)

bwplotka:
Remove explicit schema URL when building tracer provider resource in `tracing/tracing.go` to avoid schema URL mismatch errors when resources are merged.

Rationale -- we don't need to have extra upgrade step: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2015#discussion_r3650551618

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -143,7 +143,6 @@ func buildTracerProvider(ctx context.Context, tracingCfg config.TracingConfig) (
 	// Create a resource describing the service and the runtime.
 	res, err := resource.New(
 		ctx,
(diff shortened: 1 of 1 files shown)


bwplotka:
Hm, can I really remove if we have explicit svcNames? 

https://github.com/GoogleCloudPlatform/prometheus/blob/cccd2a689446a113bf66c31ef51c112063279e18/tracing/tracing.go#L146

Consulted with @dashpole - should be fine as this is the source of conflict.

bernot-dev (approved):

PR #351, opened 2026-08-04
tracing: remove explicit schema URL when building tracer provider resource (archive/release-2.43.1-gmp)

bwplotka:
Remove explicit schema URL when building tracer provider resource in `tracing/tracing.go` to avoid schema URL mismatch errors when resources are merged.


bwplotka:
Wrong branch

PR #365, opened 2026-09-02
google patch[deps]: fix release-3.13.0-gmp vulnerabilities

avilevy18:
Updating Go and image vulnerabilities using
```
./gmpctl.sh vulnfix
```


bwplotka (approved):
Hope 1.26 move will not create further issues. 3.x is fine, but I'm worried on other fronts.

PR #376, opened 2026-09-18
google patch[deps]: bump golang to 1.26.8, gke-distroless/libc to 202…

pintohutch:
…60918.00_p0, and otel to v1.45.0


bwplotka (approved):
Thanks!

